PR #24, opened 2021-06-16
Option to not overwrite existing results folder

RuaridhMacd:
GenX currently overwrites all results in the specified outpath directory. This PR creates a option for GenX to create a different results directory with an automatically generated name. Having both options has been useful for me when automating many simulations for comparisons.

Created a new setting in genx_settings.yml: "OverwriteResults". GenX will overwrite existing results (the current behaviour) if this setting is missing or set to 1. Otherwise, if the named results directory already exists, GenX finds a new similar directory name (by appending _1, _2, _3, ... to the original) and writes the results there instead.

Tested successfully with the smallNewEngland cases and worked as expected with the "OverwriteResults" setting set to 0, 1, 10, or omitted from genx_settings.yml.


sambuddhac (approved):
Reveiwed, tested. Works good !!!

sambuddhac:
Reviewed and tested. Works satisfactory.

PR #29, opened 2021-06-28
Replace the MacOrWindows setting with direct checks

RuaridhMacd:
Replace the MacOrWindows setting with direct checks of what kind of OS the system is running.

The current MacOrWindows flag is easy to miss when downloading GenX examples and can cause a simulation to fail late in the run. This change removes it in favour of having Julia check the operating system directly. Tested successfully with the NewEngland cases.

The check happens at runtime (I believe). We could change it to run at compile time using the @static macro. We could also allow the user to override the choice, but I'm not sure when that would be helpful.


sambuddhac:
This is a very useful PR. So, does this mean, we can effectively just get rid of the setup["Mac or Windows"] settings parameter?

RuaridhMacd:
Yes, I believe so. If we wanted to be more thorough we could also add conditions for bsd and freebsd ([details here](https://docs.julialang.org/en/v1/manual/handling-operating-system-variation/)).

Ideally we'd turn this into a function which is called in each of the three files so any future changes only need to be made in one place. I didn't do that for now because load_... and write_... were in different folders and I wasn't sure where utility functions like that should go.

sambuddhac (approved):
Reviewed, tested, verified, and approved.

sambuddhac:
Reviewed, tested, verified, and approved.

PR #37, opened 2021-07-20
Missing eTotalCVarIn w/o storage

dharik13:
Added condition in response one issue on github to check for whether storage is present or not when adding the term eTotalCVarIn to the total variable cost in write_costs.jl


sambuddhac (approved):
Reviewed and verified

sambuddhac:
Reviewed and verified

PR #39, opened 2021-07-26
Dev

sambuddhac:
The present pull request attempts to upgrade GenX so that it can run with Julia versions 1.5.x and 1.6.x series with the latest set of environment package dependencies. 

These are the salient features of this PR:

1. A complete refurbishing of Project.toml (and associated Manifest.toml) has been done reflecting the compatible set of versions and the SHA1 codes for each of the dependencies and also the julenv.jl file has been updated.
2. Several syntax changes has been done for the DataFrames and CSV functions in the load_inputs and write_ouput files, specifically for adding new columns. 
3. Please check the write_reserve_margin_revenue.jl. There's currently a row mismatch bug showing up in line 51 : This issue has been fixed
4. Modified the documentation to reflect the version upgrade. Please check for any more modifications I have probably missed out. 
5. Changed the .travis.yml file to now run unit tests (upcoming additions in near future) and the automated build process on Julia 1.5, 1.6, and the nightly build. The first two, to ensure everything works fine with the current version, and the nightly build runs to make sure we have a smooth transition when the next version of Julia is released (1.7.x series) with minimal rework. This will make the overall automated build and unit testing process significantly slower than before.
6. Got rid of NetworkExpansion parameter in GenX settings file and instead, applied Z>=2 condition wherever previously the NetworkExpansion parameter was used. Also fixed the errors in write_power_balance.jl (Please verify the results look good).
7. Ran through the build process to make sure everything is working fine (see the build report). The Codecov says it has reduced coverage. I won't worry about it too much right now, since I haven't yet written and deployed the unit tests (upcoming/next task (?))


GenXProject (changes requested):
NetworkExpansion should be included back into the model as implemented previously. It allows the user to model scenarios with fixed transmission capacities for a multi-zone system. 

Regarding the bug of not seeing netexports reported in power_balance.csv, the only change to be made is to modify the "If" condition in "write_power_balance.jl"  to check for whether the model has multiple zones or not - this implemented correctly. 


For all files except write_power_balance.jl,  please revert to original implementation that is uses an "If" condition based value of  "NetworkExpansion" flag rather than "Z>=2"

GenXProject (approved):

sambuddhac:
Version Upgrade of GenX to run on Julia 1.5.x and 1.6.x series

Issue #49, opened 2021-08-26
Remove `Network_zones` col from Network.csv?


gschivley:
The `Network.csv` file is supposed to have a column `Network_zones`. According to the [documentation](https://genxproject.github.io/GenX/dev/data_documentation/#.1.2-Network.csv) it is a mandatory column with values "Specified as z* where * is a number of the zone."

Based on [the code](https://github.com/GenXProject/GenX/blob/46a495b04061e71f58d31dfc30af1206d3cffde2/src/load_inputs/load_network_data.jl#L29) it looks like the `z*` format isn't necessary. The number of zones is specified entirely by the length of this column array.

The rest of the data in this file deals with transmission lines between zones, each of which is represented by a single row. Adding a column that is unrelated to the other data in each row is potentially confusing and a pain to create.

Is it possible to determine the number of zones based on the number of `z*` column names instead of the `Network_zones` column?

sambuddhac:
Fixed the issue by merging of PR on 08/26

PR #50, opened 2021-08-26
Network zone size from zone list

gschivley:
This PR removes the column `Network_zones` from the `Network.csv` file, as described in #49. It uses the number of `z*` columns to determine the number of regions. The documentation is modified to reflect the change.

I tested with one network file using the old format and didn't get any errors, but someone else should probably check to be sure.


sambuddhac:
Reviewed and tested. Works good !!!

PR #52, opened 2021-08-31
Minor documentation fixes

gschivley:
These are a few fixes to the data documentation file, not a complete review of all documentation.

- `Min_Cap_MW` is listed twice (second time should be `Min_Charge_Cap_MW`)
- Anything with `Demand_Flex` should be replaced with `Flexible_Demand`
- The underscore in `CO_2` doesn't have an escape backslash


sambuddhac (approved):
Reviewed and validated.

PR #62, opened 2021-09-15
Simplify readme for Method 1

cfe316:
I got rid of the part starting with "If however, the user opens a julia kernel, while not yet inside the GenX folder," ... since it was less well-documented. Perhaps this should be a separate Method.


sambuddhac (approved):
Reviewed. Looks nice

Issue #71, opened 2021-11-08
Superfluous lines in example Run.jl scripts


cfe316:
In all the example Run.jl scripts, the line `myinputs = Dict()` does nothing because it is immediately
overwritten on the next line by the `load_inputs` function. We can simplify the scripts by eliminating this line.

sambuddhac:
PR tested and merged. This issue is resolved.

PR #72, opened 2021-11-09
Simpify run scripts

cfe316:
In all the example Run.jl scripts, the line myinputs = Dict() did nothing because it is immediately
overwritten on the next line by the load_inputs function. This simplifies the scripts and reduces possible confusion as to why the line is there.

This addresses #71 .


sambuddhac:
Tested and reviewed. Works good.

PR #74, opened 2021-11-10
Fix Dimensions_graphic3 fig for use with dark mode

cfe316:
This adds a `Dimensions_graphic3_background.png` which has a white background. (I don't get rid of the version with a transparent background since it could be useful for publications.)

This allows uses with dark mode browsers to see the figure properly.


cfe316:
(Closing because I made a branching mistake)

PR #76, opened 2021-11-13
Draft PR for a thermal + storage (and fusion reactor) module

cfe316:
This draft PR is meant to be a discussion of how we propose to implement thermal plants with integrated thermal storage, and later, fusion reactors.

We plan to create a module that models a thermal generator with integrated storage. A fusion reactor is then a further variant.

First we'll discuss what we'd like to implement, then our proposed interface.

Model
=====
Pulsed fusion reactors with thermal storage are a subclass of reactors that have integrated thermal storage.
Reactors with integrated thermal storage have a heat source, a thermal storage unit, and a generator.
The heat source is not directly connected to the generator; flows must pass into the storage unit.
This is generally a good approximation since molten salt thermal storage has a very high charging efficiency.

A Generator
----------------
(such as a steam Rankine cycle, sCO2 or He Brayton cycle, or variants thereof)

* Existing/Min/Max capacities in a given node
* Investment cost per MWyr
* F o&m cost per MWyr
* Var o&m cost per MWh
* Up and down ramp rates, minimum downtime and/or uptime,
* A minimum power
* A start cost (in dollars)
* No 'fuel' (that is handled by the heat source)
* No Start fuel (May need to correct this)
* A thermal-to-electric efficiency

A Thermal Storage Unit
----------------------
* Existing/Min/Max capacities (in hours?)
* Investment cost per MWhyr
* F o&m cost per MWhyr (likely zero)
* Var o&m cost per MWhyr (likely zero)
* A charging effiency (likely 1)
* A max duration
* A min duration (?)

A Heat Source
-------------
* Existing/Min/Max capacities in a given node (?) (how to put this in lockstep with generator and storage?)
* Investment cost per MWyr (of heat output capacity)
* F o&m cost per MWyr (of heat output capacity)
* Var o&m cost per MWh (of heat output)
* Up and down ramp rates, minimum downtime and/or uptime,
* A minimum power
* A cap_size
* A start_cost (in dollars)

A Fusion Reactor
------------------------
Is a type of Heat Source.
It additionally has:

* A start_energy (this much electricity is required to start a pulse)
* A start_power (the grid must have this *power* available during the time that a pulse starts)
* A pulse_downtime (in hours or fractions of an hour)
* A pulse_max_uptime (in hours or fraction of an hour)
* recirc_warm (recirculating *electric* power when not on)
* recirc_hot (recirculating power when on)

We'll need to think about how to treat pulse uptimes of less than an hour.

Proposed Interface
==============

Inputs
--------
We propose adding one new (optional) integer-valued column to `Generators_data.csv` and one new file called `thermstor_data.csv`.

The new column will be called `THERMSTOR` in analogy to THERM and STOR.
Most generators will have `THERMSTOR = 0`, but
Generic thermal+storage reactors will have `THERMSTOR = 1`.
Fusion reactors will use another integer, like `THERMSTOR = 9001`.

(Perhaps in a future iteration, the interface to specify a plant with integrated thermal storage would be THERM=1 and STOR=1, but for now that would seem to require further modification and logic. It's probably easier to just add a new column.)

`Generators_data.csv` will contain data relevant to the Generator itself: everything in the Generator section above.
`thermstor_data.csv` will have data relating to the Heat Source and the Thermal Storage Unit (and this would be where to put additional data for fusion reactors).

**Rejected alternate input**
Technically we could fit most of the inputs for both the TSU and Generator in `Generators_data.csv`, but then (for columns like Existing_Cap_MW) that would result in entering values with MWth in the same columns as similar values of MWe.
This way the values in `Generators_data.csv` stay in electric units.

**Counterpoint:** Isn't the difference between MWth of storage and MWe of storage accounted for by the Eff_Dn for storage units?

Outputs
----------
We'll want an output for each zone and (if using integer unit commitment (?)) for each plant.
* Build data: Capacities of generators, storage, and reactors
* Generator's data: start and stop times, warm/hot status, power produced
* TSU's data: State of charge, charge rate, discharge rate
* Heat source's data: Start and stop times, warm/hot status, power produced
* Additional fusion statistics: recirculating energy consumed, total number of pulses.

TSU charging data should probably *not* go into `charge.csv` since it's not taking energy from the grid.
Generator power, but not heat source power, should go into `power.csv`.
May need a new file for the Heat Source data.

Implementation
============
We've not yet discussed an implementation in detail.


ASamarkRoth:
We would be very interested in this feature of GenX. Why was the PR closed?

Best, 
Anton and Co.

Issue #79, opened 2021-12-01
Loading of files can be simplified by using joinpath


cfe316:
By constructing paths to CSV files using the built-in function `joinpath`  (which returns a `String)` we can remove the `sep` variable and its being passed as an argument to each `load_xxx` function. This makes the code simpler.

gschivley:
Looks like this is now addressed in #97 (but I don't have permission to link).

cfe316:
Yup, I forgot to link them.

cfe316:
This is also addressed by #104 .

PR #80, opened 2021-12-01
Simplify loading scripts: use joinpath

cfe316:
This simplifies the code by removing an argument 'sep' from the loading scripts. Let julia do
its job by figuring out how to join paths.

Also remove replaced/commented-out code (CSV.read calls that have been supplanted)

This addresses #79 .

(The same treatment should be applied to the output files.)


cfe316:
I made a new PR which I think should work better

PR #90, opened 2021-12-16
Add a default for MultiStage to settings

### LINKED ISSUE #89: MultiStage has no default setting
On the Dev branch, this causes runs to fail if they were not configured with MultiStage in mind. There should be a default setting in `configure_settings.jl`, probably `0`.

cfe316:
The new `MultiStage` setting had no default, so if cases did not previously have it set in the `genx_settings.yml` file they would fail. This sets a default equal to 0 (for single-stage GenX).

This fixes #89 .

I also wrote a short function to make the rest of the `configure_settings` function easier to read and/or modify.
In the future this could, for example, be used to announce/log that it default setting choices are being made.


cfe316:
Hm. I realized that with the typical `Run.jl` scripts, the `configure_settings` function isn't ever even called.
Should we incorporate it? Or get rid of it and make it more useful?

PR #97, opened 2022-01-10
Simplify loading scripts: use joinpath

cfe316:
This simplifies the code by removing an argument 'sep' from the loading scripts. Let julia do
its job by figuring out how to join paths.

Also remove replaced/commented-out code (CSV.read calls that have been supplanted)


JesseJenkins (approved):
Reviewed and approved.

PR #104, opened 2022-01-16
Use joinpath rather than passing string separator

cfe316:
This results in simpler method signatures.

Also, remove 'Mac or windows' option from the `examples/Settings/genx_settings.jl` since it's unused.


JesseJenkins (approved):
Reviewed and approved.

PR #105, opened 2022-01-16
Remove service for unsupported Julia versions

cfe316:
Since we're no longer supporting v1.4 and below, remove code that's only for them.


JesseJenkins (approved):
Reviewed and approved.

PR #110, opened 2022-01-18
Shorten the .gitignore file

cfe316:
Generally this is done by ignoring whole directories rather than a list of their
contents.

I don't have the offending directories on my own system, so perhaps this should be tested by someone who does - or just let me know if you suddenly see directories or files that should be included showing up.


sambuddhac:
Thanks @cfe316 . I have reviewed your PR and have taken a deep look at it. Unfortunately, even I don't have or didn't create those lengthy lists and so, I don't have those on my local machine either. But, I took into your suggestion and applied my best judgements looking into the nature of those files. I think some of these pertain to ongoing developments by @jf-morris on DDP method and/or @nspatank on Method_of_Morris and/or @xuqingyu on Myopic runs. I would like to request  them to see if after this merge there are still issues they might be facing about the unwanted files popping up and if so to let me know. Additionally, I am also going to put up some write-up regarding keeping the project files separate from GenX cloned repo moving forth, so that this issue doesn't show up in the future.

sambuddhac:
Approved and merged !!!

cfe316:
I agree that it's best practice to keep project files outside the repo.

PR #117, opened 2022-01-22
Simplify dataframe accesses

cfe316:
Two types of simplfications have been mode. First,

`df[!, :x][y] -> df[y , :x]`

Second,
`df[a, :][:, :R_ID] -> df[a, :R_ID]`

These may save some memory and time, but more importantly make the code easier to read. This should in turn make it easier to spot errors and make changes. I can't think of any reason why these changes would lead to functional differences in the code - if they did it's because I have a hole in my understanding of DataFrames.

I've tested this on the SmallNewEngland/OneZone and the output is identical to the commit before these changes. That being said, that example does not test all the code.


cfe316:
(I'm not sure if I should be requesting that this is pulled to `main` or `Dev` or another branch.)

cfe316:
Closing because I'd like to make sure this is based off `Dev` rather than `main`, when the two seem to have diverged.

Issue #122, opened 2022-01-29
Wrong arguments in to write_transmission_flows; causes crash when NetworkExpansion=1 and Z>1


cfe316:
(This is on the `main` branch.) This one's my fault. When deleting `sep` I also deleted the argument `setup`. I didn't see this because I only tested with Z=1 until now.

cfe316:
This was fixed by #123

PR #123, opened 2022-01-29
Fix bug on main which prevents writing many outputs

cfe316:
This is to fix a bug I introduced which was recently pulled to `main`. 

This addresses #122 (Normally I would submit directly to dev).


sambuddhac:
Approved !!!

PR #125, opened 2022-01-31
A few improvements to the morris script

cfe316:
Here's a few improvements to the Morris script. I made them in an attempt to better understand the code in preparation for my own improvements.

Improvements include

* deduplicating code by using a `Dict` relating the file that a column is in and the name of the dataframe.
* breaking up long repeated lines into shorter expressions to increase readability
* making a O(n^2) algorithm more like O(n) (though this is admittedly not performance critical)
* substituting `joinpath` for custom directory forming
* replace a for loop with builtin `cumsum` for readability
* replace `eltype(x)[]` with `similar(x, 0)` which I think is better semantically?
* replace `df[!, :col][1]` with `df[1, :col]` for brevity and less syntax

I tested before-and-after these two commits against the FLECCS example with Morris enabled and the results were identical. (I subbed in a `Random.seed!(1)` to ensure the results are identical.)

### DIFF
--- a/src/additional_tools/method_of_morris.jl
+++ b/src/additional_tools/method_of_morris.jl
@@ -15,7 +15,7 @@ received this license file.  If not, see <http://www.gnu.org/licenses/>.
 """
 
 @doc raw"""
-	morris(EP::Model, path::AbstractString, setup::Dict, inputs::Dict, outpath::AbstractString, OPTIMIZER)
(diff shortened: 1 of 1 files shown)


cfe316:
I've removed the need for the ID column and added an option to read a `MethodofMorrisRandomSeed` key from the Settings file. This is completely optional but is useful for testing - with this set the random number generator will output the same result each time. The value should be a positive integer.

Issue #129, opened 2022-02-03
Period_Map is not loaded if there are no STOR_LONG_DURATION resources


cfe316:
In `load_inputs.jl`, 

```
    # Read in mapping of modeled periods to representative periods
    if setup["OperationWrapping"]==1 && !isempty(inputs["STOR_LONG_DURATION"]) && [ some other conditions ...]
        inputs = load_period_map(setup, path, inputs)
    end
```
but surely the TimeDomainReduction scripts are useful even if there are not `STOR_LONG_DURATION` resources?

Or is the idea that one would use TDR, but not the period map, so that there is explicitly *not* transfer of stored energy between periods? And thus one would turn off OperationWrapping?

Asking because the fusion module does have long duration storage capability but is not strictly part of `STOR_LONG_DURATION`, so I'd need to change this condition.

Can the second sub-condition be deleted? Is there any *harm* in using the period map if there's no LDS=1 units?

JesseJenkins:
We use time domain reduction without LDES frequently. I agree it would be
useful to report out the period map. This can be used for example in
post-processing to recreate a synthetic annual operational or price time
series by splicing the results from sample periods in the model together in
the order dictated by the period map.

Operation wrapping should basically always be used when doing sample
periods with the TDR method. If you dont do operation wrapping, the model
will not accurately reflect storage state of charge or cycling of thermal
units, as the model would choose to begin every period with a full storage
reservoir and then drain it by the end, for example, creating 'free' energy
and lowering objective function costs. Operating wrapping should still be
on when using LDES, but the LDES functionality stitches periods back
together to track changes in state of charge across the year for long
duration storage or hydropower. For all other resources, they should still
wrap from end to beginning of periods.

What is the long-duration storage functionality of the fusion model? Are
you considering very long-term thermal storage?

cfe316:
Thanks. If I understand correctly, it sounds like there's no downside, and possible upside, to removing the `&& !isempty(inputs["STOR_LONG_DURATION"])`.

> Are you considering very long-term thermal storage?

Yes, that's what we'd like to be able to model. Perhaps molten salt (solar salt) doesn't qualify as very-long-term, but in principal hot rocks could be used.

cfe316:
This was addressed by #130

PR #130, opened 2022-02-03
Refactor period map loading logic

cfe316:
Splitting this into functions makes the very long line easier to read.
It also provides a place where the smaller functions can be modified without
affecting the main function. This may be the case if, say, there is a
new resource type which is not in STOR_LONG_DURATION but nevertheless
has LDS-like capabilities which need the Period_Map.


cfe316:
This is *related* to #129 , but does not actually change any functionality.

dharik13 (approved):

PR #131, opened 2022-02-11
Resolving first-stage-only network expansion issue in multi-stage GenX

jf-morris:
These changes resolve two bugs with regard to network expansion in multi-stage GenX. First, network expansion in any given stage was limited only by the cumulative maximum flow (Line_Max_Flow_Possible_MW) and not also the single-stage maximum reinforcement (Line_Max_Reinforcement_MW) as had been intended. Second, the sets of expandable and non-expandable lines were defined improperly for stages after the first leading to all subsequent stages being denied network expansion. This pull request resolves these two issues by (1) defining a single-stage maximum reinforcement constraint in transmission_multi_stage.jl just like that in transmission.jl for single-stage GenX, and (2) redefining the EXPANSION_LINES set in configure_multi_stage_inputs.jl to reflect the inputs as intended. 
For the sake of testing, I also changed the default Line_Max_Flow_Possible_MW in the SmallNewEngland/ThreeZones_MultiStage example to be a constant value throughout the three stages which is greater than one full stage of expansion but less than three full stages. I also changed this example's multi-stage default to be DDP, rather than myopic.


xuqingyu (commented):
For each stage, a line is expandable [EXPANSION_LINES] if:
(1) there is still room for expansion (controlled by pLine_Max_Flow_Possible_MW), AND
(2) it is able to expand in this stage(controlled by pMax_Line_Reinforcement)
How about just reverse the collection by using setdiff(collect(1:L), EXPANSION_LINES) to get the NO_EXPANSION_LINES

xuqingyu (approved):

PR #138, opened 2022-02-15
Vectorize write_storagedual

cfe316:
I should have included this in my last PR as this file does not require modifying other files in order to be vectorized. However, it required a bit of debugging so I left it out. I've now checked it; it should provide an output identical to that of the non-vectorized code.


cfe316:
My mistake, this should have had only one commit attached.

PR #142, opened 2022-02-16
Add github action test workflow

gschivley:
Add a workflow that sets up Julia 1.6, builds the GenX package and runs tests.


sambuddhac (approved):
Reviewed and approved

PR #137, opened 2022-02-14
Vectorize the simpler writing procedures

cfe316:
This is "part 1" of a series of commits that vectorize the writing procedures. This work was originally done by @xuqingyu on his `Dev_FC_QX` branch, but seems too valuable to wait until that whole branch can be merged in. I've pulled out changes to these files in the `write_outputs/` directory:

```
transmission/write_transmission_flows.jl
transmission/write_transmission_losses.jl
reserves/write_reg.jl
reserves/write_rsv.jl
ucommit/write_shutdown.jl
ucommit/write_commit.jl
ucommit/write_start.jl
capacity_reserve_margin/write_capacity_value.jl
capacity_reserve_margin/write_reserve_margin.jl
write_charge.jl
write_curtailment.jl
write_nse.jl
write_power.jl
write_power_balance.jl
write_storage.jl
write_subsidy_revenue.jl
```
All these files are relatively easy to change, as they can be done one-at-a-time: none of them effect each other. Also, compared with the original commit `519862a` I've tried to keep all the whitespace and other key lines the same, so that diffs are as small as possible (and easier to review). Altogether, these changes should not change the code outputs, except in trivial ways.
* some columns now have additional digits, i.e. `0.0` instead of `0`
* minor spelling changes, `Zone` instead of `zone`.
* columns in a slightly different order in one file.

Fixes of a known bug (lack of `* inputs["omega"]` to find annual sums, which affects 2 or 3 files), are not included, in order to make it easy to verify that outputs are otherwise identical.

------------------------------
Meanwhile, these files have changes in @xuqingyu 's commit 519862a which are, I believe, all linked together (these might be separable into two groups but I'm not sure) and I have not touched them.

```
model/core/ucommit.jl
mode/core/co2.jl
model/policies/co2_credit.jl
modle/policies/co2_tax.jl
resources/storage/storage_all.jl

write_outputs/write_co2.jl
              write_co2_cap
              write_costs
              write_net_revenue
              write_outputs
              write_storagedual
```

PR #145, opened 2022-02-17
Vectorize write_costs

cfe316:
This is a single-file solution for speeding up write_costs, which for my systems is the most expensive writing procedure. On a very small test system it sped up the write time from 3s to 0.08s, so I'm looking forward to trying it out on my system where write_cost took 8200s previously.

This PR is separate but complimentary to #137 and #139. While it's the simplest, it's probably the most important of the three.

### DIFF
--- a/src/write_outputs/write_costs.jl
+++ b/src/write_outputs/write_costs.jl
@@ -27,81 +27,82 @@ function write_costs(path::AbstractString, sep::AbstractString, inputs::Dict, se
 	T = inputs["T"]     # Number of time steps (hours)
 
 	dfCost = DataFrame(Costs = ["cTotal", "cFix", "cVar", "cNSE", "cStart", "cUnmetRsv", "cNetworkExp"])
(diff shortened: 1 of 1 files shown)


cfe316 (commented):
As a start for the review of this PR, there are three separate sections here which could probably be separate commits.

The changes from lines 27-60 (original file) to lines 27-54 (new file) are one of things, which is just (1) Don't Repeat yourself and (2) grouping together the same `if setup["ParameterScale"] == 1` condition. These I would call 'cleanup' and are not a significant part of the speedup.

Lines 95-105 (original) and 92-105 (new) are the same kinds of changes, just inside the zonal loop.

The main part of this PR is lines 64-92 (orig) and 57-89 (new). 
In order to carefully vectorize this loop I moved terms and expressions out one level at a time.
All the lines in the loop over y take the form `x = x + term1 + term2 + term3`, so the first thing was to separate these out into 
```
x += term1
x += term2
x += term3
```.
At this point, it's really easy to see which terms are duplicated. The original Line 83 and 85 are identical, and Julia was computing them twice. Instead these should only be computed once.

dharik13 (approved):

PR #149, opened 2022-02-22
Update to Unit commitment minimum up time constraint

sambuddhac:
This Commit corrects the equation EP[:vCOMMIT][y,t] >= sum(EP[:vSTART][y,e] for e=(t-dfGen[!,:Up_Time][y]):t) to EP[:vCOMMIT][y,t] >= sum(EP[:vSTART][y,e] for e=(t-dfGen[!,:Up_Time][y])-1:t) since the time indexing was wrong and would let the generators be turned on for 1 hour more than the actual min up-time.

### DIFF
--- a/src/model/resources/thermal/thermal_commit.jl
+++ b/src/model/resources/thermal/thermal_commit.jl
@@ -236,7 +236,7 @@ function thermal_commit(EP::Model, inputs::Dict, Reserves::Int)
 
 		@constraints(EP, begin
 			# cUpTimeInterior: Constraint looks back over last n hours, where n = dfGen[!,:Up_Time][y]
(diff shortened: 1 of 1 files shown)


cfe316 (commented):
I think this change would make the problem worse, as _three_ hours would be required for UpTime=1.

sambuddhac:
> I think this change would make the problem worse, as _three_ hours would be required for UpTime=1.

Just corrected it.

xuqingyu:
Can I suggest to merge this change ASAP?

cfe316:
@xuqingyu my PR #151 is an alternative to this one, which also fixes the other bugs listed in #150 .

PR #155, opened 2022-02-24
Annualize outputs for reserves and ucommit

cfe316:
The period weighting ω is now taken into account for sums of reg and rsv
values, starts and shutdowns.

Co-authored-by: Qingyu Xu <xuqingyu0610@gmail.com>


sambuddhac (approved):
Reviewed and Tested, works good.

PR #157, opened 2022-02-28
Vectorize writing of outputs

cfe316:
This PR is a combination of #137, #139 ,  #145 , #147 , and #155, but target at `main` instead of `Dev` and combined into one commit to keep the history cleaner.

Co-authored-by: Qingyu Xu <xuqingyu0610@gmail.com>


sambuddhac (approved):
Reviewed and tested, Works good !!!

PR #165, opened 2022-03-15
Fuel-supply constraints

ASamarkRoth:
This PR concerns the addition of fuel-supply constraints (Minimum and maximum supply MMBTU).

A natural use-case would be applying maximum supply constraints for biofuels. However, they also provide the user further control. 

There is an example-run located in https://github.com/ASamarkRoth/GenX-1/tree/main/Example_Systems/MinMaxSupplyExample. See also: https://github.com/ASamarkRoth/GenX-1/blob/main/Example_Systems/MinMaxSupplyExample/README.md 

The constraints are added to the top of `Fuels_data.csv` for `Time_Index = -2` (Min) and `Time_Index = -1` (Max). If these rows don't exist or if they are set to a number less than 0, no constraints are added. 

I hope this is a feature that does not already exist and one that you're interested in. I am fairly new to contributing to open source projects and julia, any feedback is welcome.

Best,
Anton


cfe316:
As a further note: I like that you have constructed a very simple example with just a few types of generators. I would suggest that you can make it *even simpler* by eliminating the generators which do not end up being built, and also by shortening the length of the simulation from an entire year to just, say, 24 hours. This way the input files are significantly shorter.

ASamarkRoth:
Thanks @cfe316 for your feedback!

I have made modifications in accordance with your comments.

Concerning "hidden" entries in the `Fuels_data.csv file`: As there already is a "hidden" row (CO2), I continued these additions. It may be a good idea to move the "hidden" entries to another file, I agree. 

You can run GenX with the existing input format, i.e., without the extra rows in `Fuels_data.csv file` required to run with fuel-supply constraints. "Backwards compatibility" is kept.

cfe316:
Thanks, this looks better.
I also think this could be a bit tighter:
This
```
		if use_minmax_supply
				if Minimum_Supply_MMBTU_content[i] >= 0
					Minimum_Supply_MMBTU[i] = Minimum_Supply_MMBTU_content[i]/ModelScalingFactor 
				else
					Minimum_Supply_MMBTU[i] = Minimum_Supply_MMBTU_content[i]
				end
				if Maximum_Supply_MMBTU_content[i] >= 0
					Maximum_Supply_MMBTU[i] = Maximum_Supply_MMBTU_content[i]/ModelScalingFactor 
				else
					Maximum_Supply_MMBTU[i] = Maximum_Supply_MMBTU_content[i]
				end
			end
```
can probably just be
```
		if use_minmax_supply
			Minimum_Supply_MMBTU[i] = Minimum_Supply_MMBTU_content[i]/ModelScalingFactor 
			Maximum_Supply_MMBTU[i] = Maximum_Supply_MMBTU_content[i]/ModelScalingFactor 
		end
```
right? Even if `Minimum_Supply_MMBTU_content` is negative it's safe to divide by ModelScalingFactor.

ASamarkRoth:
Yeah that's right! 

Fixed it.

cfe316:
Hi Anton,

Thank you for submitting this PR. I've spoken with the GenX development group, and since the code is still undergoing significant development, we don't think this is a good time to merge this into the main branches.

(You can certainly point people to your customized version!)

Cheers,
Jacob

ASamarkRoth:
Hi Jacob. 

Thanks for the feedback. Keep up the good work!

Best,
Anton

Issue #173, opened 2022-03-22
Hydro_Energy_to_Power_Ratio is a required input even with no Hydro


cfe316:
This can likely be fixed by wrapping its reference in load_inputs in `if !is_empty(inputs_gen["HYDRO_RES"])` or similar.

cfe316:
This would be fixed by #174 .

PR #174, opened 2022-03-22
Gate for accessing Hydro_Energy_to_Power_Ratio

cfe316:
This prevents Hydro_Energy_to_Power_Ratio from being accessed if there are no HYDRO generators. This means that Generators_data can be simpler in cases with no hydro.


cfe316:
This addresses #173

dharik13 (approved):

PR #182, opened 2022-03-30
DDP Post-convergence extra forward pass

jf-morris:
We identified an error in the current version of DDP where starting capacities in one stage were not guaranteed to be equal to the end capacities in the previous stage. This was due to the backward pass overwriting forward pass model results. While we develop a more sustainable formulation that does not require this extra runtime, this change adds a final forward pass after successful convergence guaranteeing both feasibility and convergence criteria. Tests have shown that the inter-stage capacity mismatches have been corrected by this change.


xuqingyu (approved):
All good except for the # of steps in the comments.

Issue #63, opened 2021-09-20
Zone costs are not summed correctly


cfe316:
Possible issue in the costs.csv output: 

There seems to be an issue with adding up the zone-specific costs. When I run the SmallNewEngland/OneZone example (with all default settings, on the current `main` branch `935f19efce8afe7344efe615bf2a19c39c68883c` or on the Dev branch `aa6fd65e063ac1f3b5d7677b682671a80e43d214`) I get
```
Costs,      Total, Zone1
cTotal,    2.46e10, 1.46e10
cFix,      2.46e10, 2.46e10
cVar,       1.39e7,  1.39e7
cNSE,0.0,0.0
cStart,0.0,0.0
cUnmetRsv,0.0,-
cNetworkExp,0.0,-
```
(some decimals have been truncated and formatted to protect the innocent)
The Zone1 cost total is lower than the fixed cost, which seems wrong. It does look like the (`Total`, `cTotal`) cost is the correct summation.

Wilson was able to reproduce this output.

xuqingyu:
This is fixed, right?

cfe316:
No.

xuqingyu:
OK, check my branch, I fixed there.

xuqingyu:
but you are right, I identified the same issue and I fixed it. If my branch is too complex to implemented, it is pretty easy to fix.

cfe316:
I think this can be fixed after #157 .

xuqingyu:
True. Let's wait on that.

sambuddhac:
On to testing #157, after wrapping up #151

dharik13:
@cfe316 @sambuddhac can we incorporate this fix into the latest merge? A few more people have asked me about it and its a relatively straightforward fix

sambuddhac:
> @cfe316 @sambuddhac can we incorporate this fix into the latest merge? A few more people have asked me about it and its a relatively straightforward fix
> 

Just merged after a routine test

sambuddhac:
Just merged

cfe316:
This has now been fixed in 0.3.

PR #185, opened 2022-04-10
Fix zonal cost summations

cfe316:
This ensures that the zonal cost subtotals include non-served energy and
the charging- and energycap- related fixed costs of storage resources.

This addresses #63 .


sambuddhac:
Reviewed, tested, and approved

PR #191, opened 2022-04-27
Use joinpath rather than manual path construction from strings

cfe316:
This simplifies functions by removing the unneeded 'sep' argument and by using the julia builtin `joinpath` to construct paths with appropriate directory separators.


sambuddhac (approved):
Reviewed, tested, and approved !!!

PR #193, opened 2022-04-27
Dev

sambuddhac:
This PR aims to resolve the bug-fixes pertaining to writing output files with the paths now mentioned correctly. It also has in the load_period_map.jl file, a much clearer way of loading the period map.


JesseJenkins (approved):
Approved. But note:

In `src/load_inputs/load_inputs.jl` lines 218-221 relate to `ParameterScale` settings and appropriate scaling of CO2 emissions for start-up fuel consumption but are commented out entirely. Please review and check if this code needs to be active or removed as unnecessary now. It appears to be unnecessary and replaced by lines 212-216, in which case these commented lines should be removed.

cfe316:
Can we merge this? Other users are going as far as submitting new issues because `main` is still broken.

sambuddhac:
> Can we merge this? Other users are going as far as submitting new issues because `main` is still broken.

Just adding one small fix to the doc page and then will merge it to main @cfe316

PR #195, opened 2022-04-29
Simplify flexible demand formation

cfe316:
Results should be exactly the same. The simplification works by relocating the painfully constructed piecewise index math into easy modular arithmetic located in the function 'hoursafter'.

This does result in a different order of constraints, which causes the results to be slightly different. The best way to prove that the model is identical before & after is by examining the .lp files directly. I did this for the ISONE_Trizone example and showed that the two are identical, even after a 3 hours Max Flexible Demand Advance.


GenXProject (approved):
Code reviewed and approved !!!

PR #196, opened 2022-04-29
Simplify dataframe accesses

cfe316:
Two types of simplfications have been mode. First,

df[!, :x][y] -> df[y , :x]

Second,
df[a, :][:, :R_ID] -> df[a, :R_ID]

These may save some memory and time, but more importantly make the code easier to read. This should in turn make it easier to make changes and spot any errors.


sambuddhac (approved):
code reviewed and approved

Issue #194, opened 2022-04-28
Suggestion: using setup instead of setup["something"] in functions


xuqingyu:
Maybe instead of using setup["something"] in the function argument, we simply use setup to untify the code.
This is has an advantage that everything in the setup can be read by the function. Plus the order of the argument won't matter anymore.  It seems dangerous if order of argument differs between
when it is called in generate_model.jl 
when it is defined.

cfe316:
This is a good idea! This will make the interfaces smaller and easier to maintain in the future. I'm working on it now.

cfe316:
This was addressed by #197 and has now been merged to `main`; this issue can be closed.

PR #197, opened 2022-04-30
Simplify model generation function calls

cfe316:
This standardizes the model generation function calls by
1. passing the setup dictionary rather than passing only certain setup
   parameters. This reduces the chance of an out-of-order mistake and
   also lessens the future burden of updating documentation.
2. Renaming
   EP = myfunction(EP...) to myfunction!(EP ...). The latter follows
   julia convention that a function which modifies its arguments ends in
   a !. It also demonstrates that returning EP was unneccessary.
   
 This addresses #194 .


sambuddhac (approved):
Code reviewed and approved

PR #206, opened 2022-05-10
Modifications in thermal_no_commit and versioned doc pages

sambuddhac:
Modified the name of the function thermal_no_commit by adding a ! in order to allow for modification of the model, Updated version in Project.toml, and modification in make.jl to allow for versioned doc pages


xuqingyu (commented):
We still need Gurobi in the project toml, if someone want to comment them back, right?

sambuddhac:
> We still need Gurobi in the project toml, if someone want to comment them back, right?

You are right, @xuqingyu . That's precisely why we have the julenv.jl file with the commented out lines for both Gurobi and CPLEX. As of now, for having those commercial dependencies, the easiest way is to run GenX through Method 2 mentioned in the README and doc pages. However, changing the root level Project.toml is a bit cumbersome for the open sourced version. Of course, any user has the freedom of checking the [deps] and [compat] for Gurobi and/or CPLEX by first running through Method 2 and manually including those in the root level Project.toml (in you want to execute GenX following method 1). BUt, then running through Method 2 (after uncommenting the respective lines of julenv.jl) is the best. I will put up this note in the revised documentation.

xuqingyu:
Personally I found using method 1 is much easier than method 2 because the method 2 is project specific. If I have thousand of cases to run but I don't want to put these cases in the GenX folder, method 1 seems to be the only option. 
Back to the topic, I believe instead of deleting these gurobi related lines in the project toml, keeping and commenting them out is better.

sambuddhac:
> Personally I found using method 1 is much easier than method 2 because the method 2 is project specific. If I have thousand of cases to run but I don't want to put these cases in the GenX folder, method 1 seems to be the only option. Back to the topic, I believe instead of deleting these gurobi related lines in the project toml, keeping and commenting them out is better.

I can certainly do it with a revised PR. But, the problem is, with any fresh run of creating the environment (if for instance, you want to update any changed version of the dependencies) based off of a julenv.jl that has the commercial dependencies commented out, the entire Project.toml will get rewritten (and therefore the commercial dependencies gone !!!). But, not a problem in those cases, where the environment already exists ... I will, in any case, go ahead and put the commented lines in there for the commercial dependencies. Thanks !!! :D

xuqingyu:
OK, let me know if you are done. Then I will approve the PR.

xuqingyu (approved):
Note that "Gurobi" lines need to be added back.

sambuddhac:
@xuqingyu I just modified the README and doc page documentation a bit and included a new file `Guide_to_Project.toml` for having all the Gurobi and CPLEX related UUIDs and versions. I would admit this isn't the most robust way of doing things but, for the moment, serves the purpose.

PR #213, opened 2022-05-19
Permit OrdinaryDiffEq 6

YingboMa:


xuqingyu:
After the workflow is finished, we should also move this change to Guite_to_Project.toml.

sambuddhac:
The build is still failing on the Julia nightly build for the DiffEq package. I suspect we gotta take a closer look at the compatible version.

xuqingyu:
@sambuddhac at least it works for me on della. It looks like a documentation error. Please solve this ASAP, otherwise the GenX 0.3.1 is not functional.

cfe316:
Closing because this is no longer relevant.

PR #223, opened 2022-06-22
Remove extraneous column Line_Min_Flow_MW

cfe316:
Remove the column from the two examples where it was listed in `Network.csv`.


xuqingyu (approved):
I agree that we should them out now. We might want them back in DCOPF work of Sam.

PR #227, opened 2022-07-01
Fixed #222 by re-scaling network_expansion.csv

### LINKED ISSUE #222: network_expansion's output should be in MW and $, not GW and M$
This is a simple `ModelScalingFactor` fix. It's confusing for the outputs to be in different units from the inputs. 

The output should not depend on whether ParameterScale is used.

cfe316:
Outputs are now in MW and $, not the internal units of GW and M$.


xuqingyu (commented):
why don't we just use modelscalefactor?

cfe316:
You're right, probably we should use the ModelScaleFactor for now. However, I think we should, in the future, separate ModelScaleFactor into PowerScaleFactor (or EnergyScaleFactor) and CurrencyScaleFactor, since they're units with two distinct 'dimensions'.

xuqingyu (approved):

PR #234, opened 2022-07-18
CRM zone membership fix

xuqingyu:
This is the suggestion made by Gunther, William R. <guntherw@coned.com>. @sambuddhac


xuqingyu:
@cfe316 this is meant to include the zone that require a negative capacity reserve margin. E.g., NYISO only require the capacity >= 80% of the NYC peak load.

xuqingyu:
We still have the restriction that if Capres_x = 0, the zone is not participating in the constraint. This is related to the NSE part of the constraint. Indeed, it doesn't really matter for the final constraint which calculates the capacity demand because the constraint is in the ≥ form. But it matters for the NSE participation.

PR #236, opened 2022-08-10
GenX Patch 0.3.1 Upgrade: Addition of HiGHS Solver, DDP bug-fixes, and documentation upgrade

sambuddhac:
This PR attempts to bring in the following changes to the main branch and make it ready for publishing the patch version 0.3.1:
1. Adds the open source solver HiGHS as the new default solver
2. Fixes the previously faced bugs on DDP multi-staged expansion model
3. Attempts to create versioned doc-pages with a more sectionalized introduction (just like the other parts of the doc-pages (Able to check its functionality only after being merged with main)
4. Updated and upgraded all the dependency versions to the latest compatible ones in the environment as well as updated doc-pages to account for the new changes
5. Tests buils only on Julia 1.6.x on the CI suite and disabled nightly until further tests and bug-fixes are performed


cfe316:
There are some big problems with this patch.
1. The Retrofit update as current written breaks backward compatibility. Therefore this would have to be called 0.4.0.
It breaks backward compat by requiring a RETRO column in `Generators_data`, which is not in any of the example files. We should fix this to make it backward compatible.

2. Additions to the Run.jl script in every example directory will cause it to break if time_domain_reduction.yml is not supplied, even when TimeDomainReduction = 0.
(Maintaining every Run.jl script separately is a large burden and is a pretty strong reason to functionalize it)

3. There is an example with multiple TDR_Results subfolders. I think the Retrofits example?

sambuddhac:
@cfe316 ... can you please review and approve (if everything looks good)? I think, now this is ready? Many thanks !!!

cfe316 (approved):

cfe316:
Good! I have not tried to use or understand the retrofit functionality, but it should not now break backward compatibility.

sambuddhac:
Tested, Merged to main

PR #252, opened 2022-08-11
Move content of Run*.jl scripts into a function

cfe316:
This greatly simplifies maintenance of run scripts. Now we only need modify one instead of many.

Simply call `run_genx_case!(path)` with `path` being any folder which is a genX case, whether multistage or not.


sambuddhac:
I need a bit more info on how to run `run_genx_case!(path)` ... What folder level I should run it from? Should I open the Julia prompt? Or, can it be run from the terminal by typing something like `julia run_genx_case()` .... Imagine I am a first time GenX and Julia user. I will accordingly update the doc page and README instructions on how to run also.

sambuddhac (approved):

PR #269, opened 2022-09-01
Fix indexing in setting eTotalCapCharge max and min

RuaridhMacd:
GenX doesn't currently respect the eTotalCapCharge  < Max_Charge_Cap_MW constraint. I've been shown cases where an asymmetric storage technology has the correct max discharge capacity and energy capacity, but not charge capacity.

I'm pretty sure it's because of this indexing error. It doesn't match any of the other indexing for similar constraints. Currently waiting on longer tests to confirm the fix.


cfe316:
It looks like the `eTotalCapCharge` was changed to `eTotalCapEnergy`, which is defined in `investment_energy.jl`. Is this intentional?

RuaridhMacd:
Oops. No, that should be eTotalCapCharge. Just a tired brain and bad copy pasting.

We've tested the system with a STOR==2 technology and setting the constraint on the eTotalCapCharge[y] expression and GenX is now respecting the eTotalCapCharge < Max_Charge_Cap_MW constraint properly

cfe316 (approved):
This should fix the issue.

PR #278, opened 2022-09-30
Time Varying Startup Cost

sambuddhac:
This PR attempts to modify the startup cost formulation in commit.jl to reflect the variation with the hours. Instead of the original formulation:

`@expression(EP, eCStart[y in COMMIT, t=1:T],(inputs["omega"][t]*inputs["C_Start"][y]*vSTART[y,t]))`

it's now written as

`@expression(EP, eCStart[y in COMMIT, t=1:T],(inputs["omega"][t]*inputs["C_Start"][y,t]*vSTART[y,t]))`

And the doctoring has also been adjusted.


JesseJenkins (approved):
Approved

PR #282, opened 2022-10-06
Gotten rid of OrdinaryDiffEq from Project.toml

sambuddhac:
Currently there's no need of this dependency and it ends up creating issues. So, better to get rid of it for now.


JesseJenkins (approved):
Approved.

PR #287, opened 2022-10-21
Added docs.yml to enable doc page versioning

sambuddhac:
This PR attempts to enable doc page versioning. The reason I am targeting the main branch is due to the fact that doc pages are rendered only from the main. But, if any of the reviewers thinks that this would cause serious disruption to a linear commit history, please advise to change it to Develop (?)


JesseJenkins (approved):
Approved

PR #296, opened 2022-11-04
Deleting doc.yml

sambuddhac:
GenX/.github/workflows/docs.yml it seems, presently causes the build to fail for Julia 1.9 as well as the documentation build. So, deleting this. This was originally meant for doc page versioning.


cfe316 (approved):
Approving in order to try more things

Issue #288, opened 2022-10-21
Input file names are inconsistently case-sensitive depending on operating system


RuaridhMacd:
Some users have run into problems when using lower-case names for their input files. This does not usually cause an error on MacOS or Windows, which do not have case-sensitive filesystems, but can cause a variety of problems on Linux, which does. There are also some edge-cases where MacOS and Windows will have issues.

This could be addressed in a variety of ways. Some options are:
- Make it very explicit in the documentation that the file names are case-sensitive
- Add a config file which defines the input filenames, so they're no longer hard-coded.
- Have GenX look for lower case and capitalized versions of the input file names.

cfe316:
The third option here is easiest - it'll look for files in a case insensitive manner and then send an error message if nothing is found. 

This does mean that we won't intentionally be allowed to have both  `Generators.csv` and `generators.csv` but that would be pretty ugly anyway. I can look at this after #294 is pulled since it's somewhat overlapping with that.

RuaridhMacd:
That sounds good. I think just searching for both would be easiest, rather than trying to get fancy based on the OS. If the folder has both it'll be caught early and the error will be easy for the user to digest.

sambuddhac:
This issue is fixed now with the merge of PR#300 onto `Develop` branch.

sambuddhac:
Now fixed on main as well.

PR #300, opened 2022-11-08
Functionalize Dataframe loading from csv

cfe316:
Add a function to load a dataframe from our common sort of CSV file. This helps DRY with regard to dataframe loading settings, making code easier to read and ensuring the same options (header=1, copycols=false) are applied everywhere. Also, if we want to make dataframe loading smarter, now there's one centralized place to do it.

This commit also adds a feature that CSV files of different case can also be loaded, addressing #288 . It does not look for folders or yaml files with different cases. @RuaridhMacd it would be great if you could test this on a case-insensitive file system. 
The current implementation is sort of a minimum-working-example, it could certainly be better.

### DIFF
--- a/src/GenX.jl
+++ b/src/GenX.jl
@@ -20,6 +20,7 @@ module GenX
 export configure_settings
 export configure_solver
 export load_inputs
+export load_dataframe
 export generate_model
 export solve_model
 export write_outputs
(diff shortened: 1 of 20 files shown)


sambuddhac (approved):
I have reviewed the code modifications, tested successfully with different cases in the CSV filenames, and also pushed some further modifications for Method of Morris and Multi-Stage DDP

RuaridhMacd:
@cfe316 I won't be able to test this on a Linux immediately, but will try to do so next early week

RuaridhMacd:
Sounds good. Windows and Mac aren't case-sensitive in most cases. Apparently there are a few rare edge cases when using very long names. However, more and more of our runs are being done on clusters using Linux, so we're running into more Linux-issues.

sambuddhac:
We tested on both Mac and Windows and it works perfect. Sometimes this week/weekend, I'll test on Linux as well and let you know.

PR #303, opened 2022-11-08
Variability feature

cfe316:
A column called "None" will be automatically added in memory if it is not listed in Fuels_data.csv.

Columns of all 1.0's will be added for "non-variable" generators if they are not listed in Generators_variability.csv.


sambuddhac (approved):
Reviewed, tested, and approved !!!

PR #318, opened 2022-12-12
Adds macro to expedite model building

sambuddhac:
By disabling string name reporting to solver


cfe316 (approved):
This is fine. Please make sure to squash these 36 commits down to 1.

PR #333, opened 2023-01-14
Storage capres modifications

wilson-ricks:
Storage and reservoir hydro can now contribute to the capacity reserve margin both by directly discharging and by holding stored energy in reserve. The user must specify a new settings tag, "CapResPeriodLength", denoting the expected length of the longest period over which the capacity reserve margin constraint is binding. The model will print a message after solving, telling the user whether the chosen CapResPeriodLength was accurate.


cfe316:
It looks like this PR is mixing capacity reserve margin modifications and multi-stage model modifications. This should probably be split into two PRs.

sambuddhac (approved):
Reviewed the code and approved !!!

Issue #337, opened 2023-01-20
Problem with OperationWrapping=1 when TimeDomainReduction=0


RuaridhMacd:
We've been running some cases without TDR (TimeDomainReduction=0) but in some cases OperationWrapping was accidentally left on. This mean that GenX applied weights to the solution (i.e. inputs["omega"] != 1).

Is this the correct behaviour? It meant that the emission AnnualSum and some other calculations were wrong.

The weights were created from either pre-created TDR weights in the TDR_results folder, or the number of sub-periods within each period (inputs["H"]).

I've created a possible solution where we sanitize the inputs from bad combinations of settings in branch OpWrap-fix-with-TDR=0

RuaridhMacd:
If this seems like a bug, I can create a pull request with the fix.

cfe316:
If one doesn't already exist, it might be nice to have a function

`validatesetup!(setup::Dict)` or similar that then calls your specific function, 
`validateoperationswrappingandtdr!( .... )`

(don't take function names / capitalization / spacing style as gospel, I'm not sure about best practices here)

RuaridhMacd:
That was my thinking too. I've created a function sanitize_settings!(settings::Dict{Any,Any}) in the linked branch.

It involves some hard-coded if -> @warn -> adjust settings statements, which isn't great. I tried a version where you could define a JSON / YAML file of checks and actions but it was just as much for a developer to write out and much harder to follow

cfe316:
I think hard coded is fine here, that would be way more complication. And right now I don't think most settings are too interlinked, so each validation step should only affect one or a few settings.

For me `sanitize` suggests that it'll correct bad inputs, whereas `validate` suggests it'll error if anything is bad (and tell you how to fix it).
In some sense validation should be easier to program, but may be worse from a user perspective.
I would guess `sanitize` should only change things if there is one unambiguous way to fix it. Any ambiguity should result in halting.

cfe316:
Also, if you're game and comfortable, I'm learning the basic of Typing. We might define a `Settings` as a type alias for `Dict{AbstractString, Any}` or something like that. Then we maybe can have `sanitize!(settings::Settings)`.

RuaridhMacd:
Agreed. I leaned towards sanitizing because the error / warning messages aren't very obvious unless you're looking for them (the solver output is >50% of the output in most of my runs). I agree that we should use `error()` to kill a run if there's an ambiguous fix.

[...]

PR #346, opened 2023-02-02
Validate settings to avoid OperationWrapping=1 and TDR=0 bug - Redo for new weekly_develop branch

RuaridhMacd:
Currently, OperationWrapping=1 and TDR=0 will produce bad results, as described in issue #337 

This validates the settings and stops the run if a bad combination of settings is found. More bad combinations can be added to the function.

In the future we'll get rid of the OperationWrapping setting.


sambuddhac (approved):
Approved !!!!

RuaridhMacd:
Thanks @sambuddhac

sambuddhac:
@RuaridhMacd you're welcome. I'll put the key and lock the branch after placing the revised PRs from my end.

PR #348, opened 2023-02-08
CO2 Cap Condition check

sambuddhac:
This PR modifies the generate_model.jl file by changing the following line of code:

co2_cap!(EP, inputs, setup)

to

if setup["CO2Cap"] >= 1
		co2_cap!(EP, inputs, setup)
end
This is essential because if CO2 capacity constraints are not taken into consideration, then we don't need to execute the co2_cap!(EP, inputs, setup) function.


RuaridhMacd (approved):
Works for me.

PR #349, opened 2023-02-08
Changed crossover value to on/off

sambuddhac:
Instead of true/false according to new HiGHS+JuMP interface rule


sambuddhac:
> Looks good. Do we need to change the docs to say that Myallow_unbounded_or_infeasible now defaults to 'on'?

Can you please place a PR for this, while I work on the couple more PRs coming up soon? Thanks !!!

PR #356, opened 2023-02-13
Refactor settings & add ability to add arbitrary keys/values

cfe316:
This PR is mostly about refactoring the solver settings by eliminating duplication / excessive repetition of variable names.

Also, it adds the ability for the user to pass arbitrary key/value pairs to the solver.


cfe316:
I've also refactored the general `configure_settings` function. It should be identical in operation to its previous configuration, but it's easier to maintain. 
I've removed comment strings: a reader should instead see the documentation, e.g. `docs/data_documentation.md`.

sambuddhac (approved):

PR #360, opened 2023-02-21
Use the Fleccs branch's improved Morris script

cfe316:
This brings in several improvements to the method of morris script and interface that we made for the fleccs branch last year.

* Parameters from multiple input files can be modified. This could be useful for changing parameters from the new compound resources like fusion, thermal_storage, or VRE+Storage.
* The user no longer needs to list a copy of every generator for every parameter that needs to be changed.


sambuddhac:
@cfe316 , is this still draft or complete?

cfe316:
I needed to check out something before moving this to be ready for review.

sambuddhac:
OK !!!

cfe316:
Closing this because Filippo's implementation is better.

PR #384, opened 2023-03-15
Use a function to include all .jl files in folders

cfe316:
Remembering to add your new file to the src/GenX.jl list is an annoyance during development. This makes the include list easier to maintain by finding *all* .jl files within specified folders.

### DIFF
--- a/src/GenX.jl
+++ b/src/GenX.jl
@@ -69,152 +69,31 @@ using Cbc
 # To translate $/MWh to $M/GWh, multiply by ModelScalingFactor
 ModelScalingFactor = 1e+3
 
-# Case runner
-include("case_runners/case_runner.jl")
+# thanks, ChatGPT
+function include_all_in_folder(folder)
+    base_path = joinpath(@__DIR__, folder)
+    for (root, dirs, files) in Base.Filesystem.walkdir(base_path)
+        for file in files
+            if endswith(file, ".jl")
+                include(joinpath(root, file))
+            end
+        end
+    end
+end
 
-# Configure settings
-include("configure_settings/configure_settings.jl")
+include_all_in_folder("case_runners")
+include_all_in_folder("configure_settings")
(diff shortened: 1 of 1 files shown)


cfe316:
ChatGPT thought a better solution was to use `module`s to organize the repo, but this was the second-best option. After searching more, I don't think using many modules is that common. In any case, this is better than what we have now.

sambuddhac:
> ChatGPT thought a better solution was to use `module`s to organize the repo, but this was the second-best option. After searching more, I don't think using many modules is that common. In any case, this is better than what we have now.

Oh man !!! This is the best ZERO Lab invention of 2023, IMHO !!!

cfe316:
FYI, if there's any reason that the files (within a folder) need to be loaded in a particular order, this will likely break it. 

I could imagine that if we defined some constants in one file and then did computations with them somewhere else, this might break things. I don't know any times that we do that, other than defining ModelScalingFactor in GenX.jl itself.

sambuddhac:
> FYI, if there's any reason that the files (within a folder) need to be loaded in a particular order, this will likely break it.
> 
> I could imagine that if we defined some constants in one file and then did computations with them somewhere else, this might break things. I don't know any times that we do that, other than defining ModelScalingFactor in GenX.jl itself.

We did initially, couple years back, had this ordered way of reading the files within a folder, if I remember? But, in all fairness, you might be right that at this point, other than the model scaling factor, there's probably nothing else. But, will pay special attention to see if there's anything else.

cfe316:
> had this ordered way of reading the files within a folder, if I remember?
I'm unable to see this in the git history. :(

cfe316:
I tried reordering and then did SmallNewEngland/OneZone and costs.csv is identical. 

If we were redefining and overwriting any functions, we would get a warning. I think this is safe.

JesseJenkins (approved):

Issue #391, opened 2023-03-27
Separate New_Build and Retirement parameters


JesseJenkins:
Currently, the New_Build column in Generators_data.csv denotes whether a resource can be expanded (New_Build=1), retired (New_Build=0), or neither (New_Build=-1).

For clarity, we should seperate this into two fields each with a [0,1] binary flag:

New_Build = 1 means you can add new capacity, =0 means you cannot
Retirement = 1 means you can retire capacity, =0 means you cannot.  

Note that this will require a corresponding change in the PowerGenome scripts so please coordinate with Greg Schivley on implementation.

cfe316:
I've made it backward compatible, and using the old interface throws an informative warning.

sambuddhac:
This is now fixed.

PR #392, opened 2023-03-28
Separate new build and retirement eligibility

cfe316:
This addresses #391 .


sambuddhac:
@cfe316 , is this still draft or complete?

cfe316:
This is still pending.

sambuddhac:
OK !!!

gmantegna:
FYI this will need to be checked for compatibility with the VRESTOR module.

cfe316:
Old format to new format
------------------

New_Build[old]  to New_Build[new], Can_Retire
`1` --> `1,1`
`0` --> `0,1`
`-1` --> `0,0`

src/load_inputs/load_generators_data.jl
```diff
@@ -30,9 +30,11 @@ function load_generators_data!(setup::Dict, path::AbstractString, inputs_gen::Di
 	# Set of storage resources with symmetric charge/discharge capacity
 	inputs_gen["STOR_SYMMETRIC"] = gen_in[gen_in.STOR.==1,:R_ID]
 	# Set of storage resources with asymmetric (separte) charge/discharge capacity components
-	inputs_gen["STOR_ASYMMETRIC"] = gen_in[gen_in.STOR.==2,:R_ID]
+	STOR_ASYMMETRIC = gen_in[gen_in.STOR.==2,:R_ID]
```
gmantegna, line 33: Why do you do this for STOR_ASYMMETRIC and STOR_ALL but not STOR_SYMMETRIC? Also this is separate from the purpose of this PR, correct? That's fine if so but wanted to check.

src/load_inputs/load_generators_data.jl
```diff
@@ -95,20 +97,35 @@ function load_generators_data!(setup::Dict, path::AbstractString, inputs_gen::Di
 		inputs_gen["RSV"] = gen_in[(gen_in[!,:Rsv_Max].>0),:R_ID]
 	end
 
+    NEW_BUILD = gen_in[gen_in.New_Build.==1, :R_ID]
```
gmantegna, line 100: there seems to be some inconsistency within the code in general with tabs vs spaces. what is the standard? and can we get an auto formatter to address this as part of pre commit hooks?

src/load_inputs/load_generators_data.jl
```diff
@@ -501,4 +518,27 @@ function load_vre_stor_data!(inputs_gen::Dict, setup::Dict, path::AbstractString
...
\ No newline at end of file
+end
+
+function get_resources_which_can_be_retired(df::DataFrame)::Set{Int64}
+    if string(:Can_Retire) in names(df)
+        retirable = df[df.Can_Retire.==1, :R_ID]
+
+        invalid_newbuild = df.New_Build.==-1
+        if any(invalid_newbuild)
+            @info "Deprecated '-1' entries in the 'New_Build' column.
```
gmantegna, line 529: Suggest throwing an error (or at least strong warning) instead of an @info if users do have the Can_Retire column.

src/load_inputs/load_generators_data.jl
```diff
@@ -501,4 +518,27 @@ function load_vre_stor_data!(inputs_gen::Dict, setup::Dict, path::AbstractString
...
+had New_Build=-1 should now have New_Build=0, Can_Retire=0."
+        end
+    else
+        # Backward compatibility.
+        retirable = df[df.New_Build.!=-1, :R_ID]
+        if !isempty(retirable)
+            @info "The generators input file, 'New_Build' column, has some entries
+which are -1 (indicating the ability to be retired).  This input format is deprecated
+and may be removed in a future version. Instead, use a column 'Can_Retire',
+with entries of either 0 or 1. New_Build should be similarly restricted to {0,1}."
```
gmantegna, line 540: Can we reference the documentation here?

gmantegna (changes requested):
See comments for suggestions. I have not stress tested personally but I will once the comments are addressed.

src/load_inputs/load_generators_data.jl
cfe316: Prefer four spaces in new code. 
Yes, but let RSE do the reformat and figure out how to `git blame` past this.

src/load_inputs/load_generators_data.jl
```diff
@@ -501,4 +516,27 @@ function load_vre_stor_data!(inputs_gen::Dict, setup::Dict, path::AbstractString
...
+        if any(invalid_newbuild)
+            @warn "Deprecated '-1' entries in the 'New_Build' column.
+This previously indicated inability to build or to retire. Entries which previously
+had New_Build=-1 should now have New_Build=0, Can_Retire=0."
+        end
+    else
+        # Backward compatibility.
+        retirable = df[df.New_Build.!=-1, :R_ID]
+        if !isempty(retirable)
+            @info "The generators input file, 'New_Build' column, has some entries
```
gmantegna, line 535: This info message is not accurate. The "retirable" set which triggers the warning refers to resources with New_Build *not* equal to -1, which is inconsistent with the warning message which complains about resources with New_Build equal to -1. Perhaps we can just throw a more generic warning message about not having the Can_Retire column if this column is not found? And remind users of the correct input format? (What 0/1 means in each column)

gmantegna (changes requested):
See comment on info statement.

src/load_inputs/load_generators_data.jl
cfe316: Plz lmk if this is satisfactory.

cfe316:
Notes from the meeting today:
May want to wait until v0.4 in order to make some backward-incompatible changes.

src/load_inputs/load_generators_data.jl
gmantegna: Putting my Slack comments here for record keeping purposes: I think that we should require the Can_Retire column, and throw the info message anytime it is not present. Otherwise, it will not be possible to change the default Can_Retire value to 0 (which I would assume we would want if this column is optional) without breaking backward compatibility.

src/load_inputs/load_generators_data.jl
```diff
@@ -599,4 +616,31 @@ function validate_piecewisefuelusage(heat_rate_mat, load_point_mat)
...
+    else
+        retirable = df.Can_Retire .== 1
+    end
+    return Set(findall(retirable))
+end
+
+function resources_which_can_be_built(df::DataFrame)::Set{Int64}
+    buildable = df.New_Build .== 1
+    return Set(findall(buildable))
+end
```
cfe316, line 646: These are the key functions.

cfe316:
@gmantegna This is complete now.

Backward compatible (with warning) :heavy_check_mark: 
Documentation :heavy_check_mark: 
Tested :heavy_check_mark: 

I suppose for safety we might want to add a check that if you're using the new interface there should be only 1 or 0 in New_Build.

gmantegna (approved):
Thanks!

PR #405, opened 2023-04-14
SCIP commented out from Project.toml

sambuddhac:
SCIP doesn't load properly on Windows. Until the matter is resolved at SCIP's end, we're not using it as a solver for GenX.


cfe316 (commented):
Is there any associated documentation which needs to be changed?

sambuddhac:
> Is there any associated documentation which needs to be changed?

Good catch, @cfe316 . Let me update that too.

sambuddhac:
Updated the documentation @cfe316

cfe316:
Also might want to remove it from the doc page `solver_configuration.md`

PR #407, opened 2023-04-14
Abstracting REG and RSV terms for therm-commit ramp

cfe316:
A recent change to the THERM_COMMIT ramp expessions added calls to :vREG and :vRSV variables. However, these variables did not exist at all unless setup["Reserves"] > 1; and further they would only exist for a subset of the thermal generators. This would throw an error whenever Reserves was not 1 or if not all thermal generators with committment had positive Reg_Max and sv_Max entries.

This change introduces expressions that are defined for all generators y in THERM_COMMIT, which are 0 if Reserves are off and reference the vREG and vRSV variables for the correct generators if Reserves are on.

This lets us modify the ramp expression to include these reserve and regulation terms, while not needing multiple copies of the expression.

This technique could be used to further simplify the Reserve expressions below.

### DIFF
--- a/src/model/resources/hydro/hydro_res.jl
+++ b/src/model/resources/hydro/hydro_res.jl
@@ -86,6 +86,19 @@ function hydro_res!(EP::Model, inputs::Dict, setup::Dict)
 	HYDRO_RES = inputs["HYDRO_RES"]	# Set of all reservoir hydro resources, used for common constraints
 	HYDRO_RES_KNOWN_CAP = inputs["HYDRO_RES_KNOWN_CAP"] # Reservoir hydro resources modeled with unknown reservoir energy capacity
 
+    # These variables are used in the ramp-up and ramp-down expressions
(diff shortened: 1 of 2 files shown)


sambuddhac:
Thanks @cfe316

cfe316:
Note that here I'm using anonymous expressions, which I think is unique in GenX. This is nice because it doesn't pollute the wider namespace.

sambuddhac:
The reserves term should be  

```
reserves_term = @expression(EP, [y in THERM_COMMIT, t in 1:T],
                           y ∈ THERM_COMMIT_RSV ? EP[:vRSV][y,t] : 0)
```

instead of 

```
 reserves_term = @expression(EP, [y in THERM_COMMIT, t in 1:T],
                           y ∈ THERM_COMMIT_REG ? EP[:vRSV][y,t] : 0)
```

sambuddhac (approved):

PR #413, opened 2023-04-15
Validate the time basis

cfe316:
This should prevent user error or errors stemming from malformed PG input.

This checks that the number of representative periods and their length equal the number of steps in each of the input time series.
Also checks that the number of subperiod weights is the number of representative periods.


cfe316:
Try to get past the validator! I dare you!

sambuddhac (approved):
Code reviewed, tested on ISONE_Trizone_FullTimeSeries case with TDR on. But, eventually, this function needs to have automated testing.

PR #416, opened 2023-04-18
Fix typo in fix_initial_investments

filippopecci:
y in keys(EP_cur[c]) is a DenseAxisArray key, to extract the index we need to have y[1], as already done also on line 350 in function fix_capacity_tracking.


filippopecci:
Closing to re-open from develop

PR #418, opened 2023-04-25
Re-refactor configure_settings

cfe316:
This re-applies changes from #356 that were undone when merging.


cfe316:
Closing due to irrelevance

PR #420, opened 2023-04-25
Unset executable bits

cfe316:
At the moment, many files in the GenX tree have the 'executable' bit set. This is confusing and not actually useful, since these files are not, in fact, executable: they're all data to be used as inputs. The only executable used in GenX is the user's installation of `julia`.

This PR unsets the executable bits.


cfe316:
See [https://en.wikipedia.org/wiki/File-system_permissions#Permissions](https://en.wikipedia.org/wiki/File-system_permissions#Permissions)  or [https://www.digitalocean.com/community/tutorials/an-introduction-to-linux-permissions](https://www.digitalocean.com/community/tutorials/an-introduction-to-linux-permissions) for introductions  to this topic.

PR #424, opened 2023-04-27
Add a change log.

cfe316:


sambuddhac (commented):
Shouldn't this be targeted to the weekly_develop? I think so. Every time we publish a patch, we'll also need the release notes, right? Or, no?

cfe316:
Thanks. Yes, fixed now.

sambuddhac (approved):
Thanks !!! Approved !!!

PR #426, opened 2023-04-28
Reform the interface with TDR by eliminating the OperationWrapping flag

cfe316:
This is a WIP. 

* Remove the OperationWrapping setting.
* Always load the period weights (from `load_data.csv`) and make sure they match the provided data.
* Add a warning if a user tries to use OperationWrapping.

In some places (particular for deciding whether to activate LDS formulations),  inputs["REP_PERIOD"] > 1 is used instead of setup["OperationWrapping".


cfe316:
Still needs attention:
* Documentation.
* Validation: throw an error if time domain data has already been reduced but TDR is on.
* Testing: making sure the LDS formulation is correct.

src/case_runners/case_runner.jl
```diff
@@ -92,8 +93,11 @@ function run_genx_case_multistage!(case::AbstractString, mysetup::Dict)
     ### Cluster time series inputs if necessary and if specified by the user
     tdr_settings = get_settings_path(case, "time_domain_reduction_settings.yml") # Multi stage settings YAML file path
     TDRSettingsDict = YAML.load(open(tdr_settings))
-    TDRpath = joinpath(case, "Inputs", "Inputs_p1", mysetup["TimeDomainReductionFolder"])
+
+    first_stage_path = joinpath(case, "Inputs", "Inputs_p1")
+    TDRpath = joinpath(first_stage_path, mysetup["TimeDomainReductionFolder"])
     if mysetup["TimeDomainReduction"] == 1
+        prevent_doubled_timedomainreduction(first_stage_path)
```
cfe316, line 100: Here, we only check the configuration of the first stage of a multi-stage problem.
The other stages are checked in the `cluster_inputs` function.

src/load_inputs/load_inputs.jl
```diff
@@ -63,7 +63,7 @@ function load_inputs(setup::Dict,path::AbstractString)
 	end
 
 	# Read in mapping of modeled periods to representative periods
-	if is_period_map_necessary(setup, path, inputs) && is_period_map_exist(setup, path, inputs)
+	if is_period_map_necessary(inputs) && is_period_map_exist(setup, path, inputs)
```
cfe316, line 66: I slightly changed the interface here as the other args were not necessary.

src/load_inputs/load_inputs.jl
```diff
@@ -72,11 +72,11 @@ function load_inputs(setup::Dict,path::AbstractString)
 	return inputs
 end
 
-function is_period_map_necessary(setup::Dict, path::AbstractString, inputs::Dict)
-	ow = setup["OperationWrapping"]==1
+function is_period_map_necessary(inputs::Dict)
+	multiple_rep_periods = inputs["REP_PERIOD"] > 1
```
cfe316, line 76: Generally, OperationWrapping == 1 was used to mean that there are multiple representative periods. In several places I change this to the more obvious check.

src/load_inputs/load_load_data.jl
```diff
@@ -28,28 +28,20 @@ function load_load_data!(setup::Dict, path::AbstractString, inputs::Dict)
...
+    inputs["REP_PERIOD"] = convert(Int16, as_vector(:Rep_Periods)[1])
+    inputs["H"] = convert(Int64, as_vector(:Timesteps_per_Rep_Period)[1])
+
+    # Creating sub-period weights from weekly weights
+    for w in 1:inputs["REP_PERIOD"]
+        for h in 1:inputs["H"]
+            t = inputs["H"]*(w-1)+h
+            inputs["omega"][t] = inputs["Weights"][w]/inputs["H"]
+        end
+    end
```
cfe316, line 44: The subperiod weights are now *always* loaded.

src/load_inputs/load_load_data.jl
```diff
@@ -142,3 +134,31 @@ function validatetimebasis(inputs::Dict)
...
+        error("""Critical error in time series construction:
+              Time domain reduction (clustering) is being called for,
+              on data which may already be clustered. In demand_data.csv [or load_data.csv],
+              the number of representative periods (:Rep_Period) is ($representative_periods)
+              and the number of subperiod weight entries (:Sub_Weights) is ($num_sub_weights).
+              Each of these must be 1: only a single period can have TimeDomainReduction applied.""")
+    end
+
+end
+
```
cfe316, line 163: This check might not be strictly necessary: `cluster_inputs` might fail with bad inputs. But this at least gives a more informative message.

src/write_outputs/write_outputs.jl
```diff
@@ -108,7 +108,7 @@ function write_outputs(EP::Model, path::AbstractString, setup::Dict, inputs::Dic
 
 
 	# Output additional variables related inter-period energy transfer via storage
-	if setup["OperationWrapping"] == 1 && !isempty(inputs["STOR_LONG_DURATION"])
+	if !isempty(inputs["STOR_LONG_DURATION"])
```
cfe316, line 111: If there's only one representative period, STOR_LONG_DURATION will be empty.

docs/src/methods.md
```diff
@@ -1,7 +1,87 @@
...
+julia> run_timedomainreduction!("/path/to/case")
+```
+
+This function will obey the settings in `path/to/case/Settings/time_domain_reduction_settings.yml`.
+It will output the resulting clustered time series files in the case.
+
+Running this function will *overwrite* these files in the case.
+This is done with the expectation that the user is trying out various settings.
+
+
```
cfe316, line 83: This section was moved from the main `how_to_run_genx.md` page.

cfe316:
I've tested this on the RealSystemExample/ISONE_Singlezone, which has TDR and LDS, and the results were identical before and after.

cfe316:
Also - please "squash and merge" when this is ready to be pulled in. Or just let me know and I'll smush it down to a single commit.

src/write_outputs/write_outputs.jl
dharik13: Make sense.

CHANGELOG.md
```diff
@@ -9,12 +9,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
-- This changelog
+- Added a description of how to use time domain reduction (#426).
+- Validation: against trying to perform time domain reduction (clustering)
+  on data which has already been clustered.
+- This changelog (#424).
 
 ### Fixed
 
 - Not having a changelog (#423).
```
RuaridhMacd, line 19: Lol

Example_Systems/RealSystemExample/ISONE_Singlezone/README.md
RuaridhMacd, line 1: More of a style question for @JesseJenkins and @dharik13 , but I assume most users run GenX from an interactive session, not the command line, so might it be better to keep _include("Run.jl")_ as the way of running optimizations? I think the current explanation of julia --project=. and then running that way is well explained and understood.

src/load_inputs/load_inputs.jl
RuaridhMacd: The function looks good, but should we avoid implicit returns? I generally find it only leads to confusion and doesn't play as nicely with some VSCode plug-ins.

RuaridhMacd (approved):
The process of making and using TDR inputs is much simplified and should hopefully avoid common user errors that we've seen.

The various checks, e.g. prevent_doubled_timedomainreduction, worked on the examples I tried.

src/load_inputs/load_inputs.jl
cfe316: I'm neutral on this.
Pro implicit returns: "don't have code which does nothing / is not necessary" "minimalism"
Con: Potentially, worse readability. Doesn't work with VS code plugins.

I'd support an overall code-style-consistency rewrite, with something like `black`.

PR #430, opened 2023-05-01
Remove TDR_Results folders from examples

cfe316:
Also remove a wayward Project.toml file and a random lock file.


sambuddhac:
Is the removal of TDR_Results folder to enable time domain clustering script to run instead of referring to existing tdr result, which may contain a different set of weights and/or set of rep periods than what's desired by some particular user?

sambuddhac (commented):
How about including the TDR_Results in the .gitignore, rather?

sambuddhac (dismissed):
Can we move the TDR_Results folder to .gitignore rather?

cfe316:
> Is the removal of TDR_Results folder to enable time domain clustering script to run instead of referring to existing tdr result, which may contain a different set of weights and/or set of rep periods than what's desired by some particular user?

The TDR_Results are an *output* of the code, and we should not be in the business of storing outputs.
As an example which is focused on the multistage outputs and not specifically on the TDR outputs, I expect that the user will not need to mess with the time domain reduction settings.

A line to ignore TDR_Results is already in the .gitignore file, but these particular folders were in the repo at the time of the addition of that line, so they're still in the repo now.  This PR removes them, so that users on the latest commit won't see them as being pre-existing.

sambuddhac:
Thanks. It totally makes sense to me now.

sambuddhac (approved):

PR #431, opened 2023-05-01
Remove superfluous call to `haskey`

cfe316:
Because of the new default settings, the first term in these expression always evaluates to false. They can be removed.


sambuddhac (approved):
Approved

PR #436, opened 2023-05-08
Update make.jl to update the documentation pages for GenX v0.3.5

sambuddhac:
Doc page upgrade: moving the input output before additional tools, fleshing out the multistage, getting rid of/adding content to the missing content in doc pages.
Better integrating multi-stage expansion instructions (there should be more clarity about the three ways to run GenX: single stage, multi-stage myopic, and multi-stage foresight with discussion on what that means for computational performance and how input files change for each mode)

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -50,13 +50,8 @@ pages = OrderedDict(
             ]
         ],
         "Multi_stage" => [
-            "Overview" => "multi_stage_model_overview.md",
-            "Load inputs for multi-stage modeling" => [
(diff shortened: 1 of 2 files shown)


cfe316 (approved):
We've reordered two entries and lightly edited some of the data documentation.

cfe316 (approved):

cfe316 (changes requested):
This commit says it involves better description of the various ways to run GenX, but there's not anything about this?

PR #442, opened 2023-05-12
Update configure_settings.jl, pre cluster.jl and time_domain_reductio…

sambuddhac:
…n.jl for multistage TDR bug fix

Needs updating of the way multi_stage_settings.yml is read and stored.


sambuddhac:
Hi @cfe316 , seems we need a bit of a work to get this going on the `time_domain_reduction.jl` still. And probably a better refactored/optimized code too (your domain expertise)

PR #444, opened 2023-05-12
Update README.md and doc pages for version bump to v0.3.5 and environ…

sambuddhac:
…ment upgrade


Project.toml
```diff
@@ -27,7 +27,7 @@ YAML = "ddb6d928-2868-570f-bddf-ab3f9cf99eb6"
 CSV = "0.10.4"
 Cbc = "1.0.1"
 Clp = "1.0.1"
-Clustering = "0.14.2, 0.15"
+Clustering = "0.15"
```
cfe316, line 30: Why was this changed down to only support 0.15?

Project.toml
```diff
@@ -37,7 +37,7 @@ JuMP = "1.1.1"
 MathOptInterface = "1.6.1"
 RecursiveArrayTools = "2.31.2"
 Statistics = "1.4.0"
-StatsBase = "0.33.21, 0.34"
+StatsBase = "0.34"
```
cfe316, line 40: Why was this changed to only support 0.34?

cfe316 (commented):
This is fine, but I'm a bit confused why the dependencies became more restricted. There shouldn't be a reason to restrict them unless we're all of a sudden using a new function which is not available in the old version.

PR #452, opened 2023-05-17
Update CompatHelper.yml

sambuddhac:
To set the frequency of environment upgrade to monthly


cfe316 (approved):
This is good; let's try reducing the frequency of these requests. Now it'll propose changes just once per month.

PR #465, opened 2023-05-24
fix bug related to writing LDES outputs when you are not using TDR

gmantegna:
Currently, if you are modeling long duration storage but not using any time domain reduction, GenX will encounter an error when writing the outputs, because the data related to TDR (i.e. PeriodMap) does not exist. This changes fixes the issue by only printing LDES specific outputs when TDR is being used. NOTE: this means that LDES outputs will not be printed if the user is inputting data with TDR already performed (is there a better way to do this?).


cfe316 (changes requested):
<finishing this review so that it's visible to Gabe>

PR #467, opened 2023-05-24
Handle Pre_Solve key for CPLEX

cfe316:
This allows CPLEX to handle the Pre_Solve key which is in many examples.
Note that 1 is the default, which is to have presolve turned on.


sambuddhac (approved):
Approved !!!

PR #471, opened 2023-05-29
LDES outputs bug fix + add duals outputting + storage capres

gmantegna:


src/GenX.jl
```diff
@@ -39,7 +39,7 @@ using Statistics
 
 # Uncomment if Gurobi or CPLEX active license and installations are there and the user intends to use either of them
 #using CPLEX
-#using Gurobi
+using Gurobi
```
cfe316, line 42: Please remove this line - it'll cause trouble for anyone trying to build. Same with the add to Project.toml.

src/model/resources/storage/storage_all.jl
```diff
@@ -9,6 +9,7 @@ function storage_all!(EP::Model, inputs::Dict, setup::Dict)
 
 	dfGen = inputs["dfGen"]
 	Reserves = setup["Reserves"]
+	CapacityReserveMargin = setup["CapacityReserveMargin"]
```
cfe316, line 12: Could you use lower case here for the variable name?

src/write_outputs/write_net_revenue.jl
```diff
@@ -32,6 +32,16 @@ function write_net_revenue(path::AbstractString, inputs::Dict, setup::Dict, EP::
...
 	end
 
+	# Add capacity constraint dual
+	dfNetRevenue[!, "Capacity_constraint_dual"] = fill(NaN, nrow(dfNetRevenue))
+	for y in intersect(dfGen[dfGen.Max_Cap_MW.>0, :R_ID], 1:G)
+		dual_val = dual.(EP[:cMaxCap][y])
+		if setup["ParameterScale"] == 1
+			dual_val *= ModelScalingFactor # converting MM$/GW-yr to $/MW-yr
+		end
+		dfNetRevenue[y, "Capacity_constraint_dual"] = dual_val
```
cfe316, line 42: Consider separating the `for` and `if`. Could you `*=` the column afterward?

Or look at the solutions in other files: above the for loop have `modelscalingfactor = setup["ParameterScale"] ? ModelScalingFactor : 1` and then always `* modelscalingfactor`.

cfe316: Is there a reason why the intersection is necessary? `R_ID` is always in the set 1:G.

cfe316 (changes requested):
I'm glad to see that you're addressing these issues.

Would it be possible to separate into different pull requests?
1 - fix bug
2- add or enhance new feature

src/write_outputs/write_net_revenue.jl
gmantegna: Yes it needs to be only the generators with a maximum capacity constraint.

PR #472, opened 2023-05-30
Fix bug in LDES outputs

gmantegna:
Previously, GenX would throw an error during outputting if any LDES technologies were being included, but time domain reduction was not being used. This change fixes that bug.


cfe316 (approved):
Gabe has tested this and found that it fixes the issue.

PR #492, opened 2023-06-23
Update README.md

sambuddhac:
With additional instructions for running from Julia REPL terminal


README.md
```diff
@@ -99,6 +101,19 @@ type `include("Example_Systems/SmallNewEngland/OneZone/Run.jl")` from the `julia
 
 After the script runs to completion, results will be written to a folder called “Results”, located in the same directory as `Run.jl`.
```
cfe316, line 102: This may be fine, but you need to be careful with newlines here as this may break the numbering. Markdown will often produce a `1.` in a numbered list even if you start from `2.`.

JesseJenkins (approved):

PR #480, opened 2023-06-09
Update write_power_balance.jl

sambuddhac:
Gotten rid of (-0.5) factor in line 43

PR #493, opened 2023-06-23
Update CHANGELOG.md

sambuddhac:
Updated with description of changes from PR 480 (Gotten rid of (-0.5) factor in line 43 of write_power_balance.jl (#480))


cfe316:
Could you please add this to #480 rather than having a separate PR? That way it can all be part of 1 commit.

PR #496, opened 2023-07-04
Combine two LDS constraints

cfe316:
Use the `mod1` function to eliminate needing separate constraints for 'interior' periods and 'start' periods.
This also removes the warning that happened because `MODELED_PERIODS_INDEX[end]` was a single value rather than vector.


cfe316:
This was tested via running on `RealSystemExample/Trizone` before and after. The objective value (cost) is identical.

sambuddhac (approved):

PR #498, opened 2023-07-11
add_to_expression!

pshlime:
In the most recent update of GenX, I, along with @RuaridhMacd , found and replaced all applicable instances of += and -= in GenX’s source code.  We replaced them with add_to_expression!, a primitive function that is called by += and -=, which is also


RuaridhMacd:
This work built off an observation @filippopecci made, that `add_to_expression!` could help the model generation step.

The runtime of `generate_model()` can go up or down a bit with this change, based on @pshlime's testing on the SmallNewEngland examples, but the memory usage is consistently less. Given that memory is more of a constraint for large models, we thought it would be a good trade-off.

@sambuddhac Let me know if you think this should be pushed to weekly_develop vs Develop. I thought this was more of a substantial change, but Develop looks a little bit behind.

src/model/core/discharge/discharge.jl
```diff
@@ -35,15 +35,20 @@ function discharge!(EP::Model, inputs::Dict, setup::Dict)
...
 
 		@expression(EP, eESRDischarge[ESR=1:inputs["nESR"]], sum(inputs["omega"][t]*dfGen[y,Symbol("ESR_$ESR")]*EP[:vP][y,t] for y=dfGen[findall(x->x>0,dfGen[!,Symbol("ESR_$ESR")]),:R_ID], t=1:T)
 						- sum(inputs["dfESR"][z,ESR]*inputs["omega"][t]*inputs["pD"][t,z] for t=1:T, z=findall(x->x>0,inputs["dfESR"][:,ESR])))
 
-		EP[:eESR] += eESRDischarge
+
+		for i=1:inputs["nESR"]
+			add_to_expression!(EP[:eESR][i], eESRDischarge[i])
+		end
+		# EP[:eESR] += eESRDischarge
```
cfe316, line 51: Can you say something about why the for loop is needed here?

src/model/core/discharge/investment_discharge.jl
```diff
@@ -139,9 +139,9 @@ function investment_discharge!(EP::Model, inputs::Dict, setup::Dict)
 		# OPEX multiplier scales fixed costs to account for multiple years between two model stages
 		# We divide by OPEXMULT since we are going to multiply the entire objective function by this term later,
 		# and we have already accounted for multiple years between stages for fixed costs.
-		EP[:eObj] += (1/inputs["OPEXMULT"])*eTotalCFix
+		add_to_expression!(EP[:eObj], (1/inputs["OPEXMULT"]), eTotalCFix)
```
cfe316, line 142: I think you can remove the excess parens here.

src/model/core/discharge/investment_discharge.jl
```diff
@@ -165,14 +165,28 @@ function investment_discharge!(EP::Model, inputs::Dict, setup::Dict)
...
+
+	x = inputs["NumberOfMinCapReqs"]
+
 	if setup["MinCapReq"] == 1
-		@expression(EP, eMinCapResInvest[mincap = 1:inputs["NumberOfMinCapReqs"]], sum(EP[:eTotalCap][y] for y in dfGen[dfGen[!, Symbol("MinCapTag_$mincap")] .== 1, :R_ID]))
-		EP[:eMinCapRes] += eMinCapResInvest
+		@expression(EP, eMinCapResInvest[mincap = 1:x], sum(EP[:eTotalCap][y] for y in dfGen[dfGen[!, Symbol("MinCapTag_$mincap")] .== 1, :R_ID]))
+		for i=1:x
+			add_to_expression!(EP[:eMinCapRes][i], eMinCapResInvest[i])
+		end
```
cfe316, line 175: Again, why is the for loop needed?

src/model/core/discharge/investment_discharge.jl
```diff
@@ -165,14 +165,28 @@ function investment_discharge!(EP::Model, inputs::Dict, setup::Dict)
 	# DEV NOTE: This constraint may be violated in some cases where Existing_Cap_MW is <= Min_Cap_MW and lead to infeasabilty
 	@constraint(EP, cMinCap[y in intersect(dfGen[dfGen.Min_Cap_MW.>0, :R_ID], 1:G)], eTotalCap[y] >= dfGen[y, :Min_Cap_MW])
 
+
+	x = inputs["NumberOfMinCapReqs"]
```
cfe316, line 169: This line potentially causes an error if NumberOfMinCapReqs is not defined. At the least, please put this inside of the `if` statement. And please don't use single-letter variable names - prefer descriptive ones like `num_mincapreqs`

src/model/core/discharge/investment_discharge.jl
```diff
@@ -165,14 +165,28 @@ function investment_discharge!(EP::Model, inputs::Dict, setup::Dict)
...
+			add_to_expression!(EP[:eMinCapRes][i], eMinCapResInvest[i])
+		end
+
+		# EP[:eMinCapRes] += eMinCapResInvest
 	end
+	# issues with running this, claims it is difference of int64 compared to AffExpr
+	# add_to_expression!(EP[:ePowerBalance][t, z], ePowerBalanceNse[t, z])
+
+	y = inputs["NumberOfMaxCapReqs"]
+
```
cfe316, line 183: Same as comment above about NumberOfMinCapReqs.

src/model/generate_model.jl
```diff
@@ -88,32 +88,74 @@ function generate_model(setup::Dict,inputs::Dict,OPTIMIZER::MOI.OptimizerWithAtt
...
 	# Initialize Power Balance Expression
 	# Expression for "baseline" power balance constraint
-	@expression(EP, ePowerBalance[t=1:T, z=1:Z], 0)
+	# @expression(EP, ePowerBalance[t=1:T, z=1:Z], 0)
+	temp = Matrix{AffExpr}(undef, T, Z)
+	for t=1:T
+		for z=1:Z
+			temp[t,z] = 0.0
+		end
+	end
```
cfe316, line 97: Surely this could be replaced with `zeros(AffExpr, T, Z)` ?

src/model/generate_model.jl
```diff
@@ -88,32 +88,74 @@ function generate_model(setup::Dict,inputs::Dict,OPTIMIZER::MOI.OptimizerWithAtt
...
+	createemptyexpression(EP, :eGenerationByZone, Z, T)
 	# Initialize Capacity Reserve Margin Expression
 	if setup["CapacityReserveMargin"] > 0
-		@expression(EP, eCapResMarBalance[res=1:inputs["NCapacityReserveMargin"], t=1:T], 0)
+		temp = Matrix{AffExpr}(undef, inputs["NCapacityReserveMargin"], T)
+		for t=1:T
+			for x=1:inputs["NCapacityReserveMargin"]
+				temp[x, t] = 0.0
+			end
+		end
```
cfe316, line 119: again, I think this can be replaced by one call to `zeros(AffExp, inputs["N..."], T)`

src/model/generate_model.jl
```diff
@@ -223,3 +265,24 @@ function generate_model(setup::Dict,inputs::Dict,OPTIMIZER::MOI.OptimizerWithAtt
...
+	EP[exprname] = temp
+end
+
+function createemptyexpression(EP::Model, exprname::Symbol, dim1::Int64, dim2::Int64)
+	temp = Matrix{AffExpr}(undef, dim1, dim2)
+	for y=1:dim2
+		for x=1:dim1
+			temp[x, y] = 0.0
+		end
+	end
```
cfe316, line 286: Again I think this can be `zeros(AffExpr, dim1, dim2)`

src/model/generate_model.jl
cfe316: or even something using your createemptyexpression function?

src/model/core/discharge/discharge.jl
cfe316: Ah, it looks like add_to_expression! only works on single values, not on matrices?
Would you consider adding a new method for it that works on matrices? Or a related function like `add_to_expression_matrix` which takes a tuple of dimensions as arguments? Or which finds the shape from the object?

cfe316 (changes requested):
Hi pshlime, this is a good first draft but I think various parts can be tightened up. Please see the comments.

src/model/core/discharge/discharge.jl
RuaridhMacd: Hey Jacob @cfe316,
We've found that add_to_expression! has some unexpected behaviours when broadcast or used in the same manner as += , -=, etc. so using nested for-loops has been a lot safer.
We could make a wrapper function but I've found using a for-loop to often be clearer and safest

gmantegna:
@pshlime I am no longer seeing the changes that this PR was intended to add in the "Files changed" section. Perhaps something went wrong when you did the recent force push? Also, please remove the changes to the example systems and the Gurobi changes from this PR.

src/model/resources/curtailable_variable_renewable/curtailable_variable_renewable.jl
```diff
@@ -39,12 +39,20 @@ function curtailable_variable_renewable!(EP::Model, inputs::Dict, setup::Dict)
 	@expression(EP, ePowerBalanceDisp[t=1:T, z=1:Z],
 	sum(EP[:vP][y,t] for y in intersect(VRE, dfGen[dfGen[!,:Zone].==z,:R_ID])))
 
-	EP[:ePowerBalance] += ePowerBalanceDisp
+	for t=1:T
+		for z=1:Z
+			add_to_expression!(EP[:ePowerBalance][t, z], ePowerBalanceDisp[t, z])
+		end
+	end
```
cfe316, line 46: Can we add methods to `add_to_expression!` that takes two integers in order to avoid an explicit 5-line for-loop?
This could ensure that the loop order (row-major vs column-major) is correct.

src/model/generate_model.jl
```diff
@@ -223,3 +225,29 @@ function generate_model(setup::Dict,inputs::Dict,OPTIMIZER::MOI.OptimizerWithAtt
...
+function createemptyexpression(EP::Model, exprname::Symbol, dims::NTuple{N, Int64}) where N
+	EP[exprname] = zeros(AffExpr, dims)
+end
+
+
+function createemptyexpression(EP::Model, exprname::Symbol, dims::Vector{Int64})
+	# The version using tuples is faster
+
+	EP[exprname] = zeros(AffExpr, dims...)
+end
```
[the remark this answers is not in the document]
RuaridhMacd: Good point. I'll change that

src/model/generate_model.jl
```diff
@@ -158,7 +160,7 @@ function generate_model(setup::Dict,inputs::Dict,OPTIMIZER::MOI.OptimizerWithAtt
 	end
 
 	# Model constraints, variables, expression related to reservoir hydropower resources with long duration storage
-	if inputs["REP_PERIOD"] > 1 && !isempty(inputs["STOR_HYDRO_LONG_DURATION"])
+	if setup["OperationWrapping"] == 1 && !isempty(inputs["STOR_HYDRO_LONG_DURATION"])
 		hydro_inter_period_linkage!(EP, inputs)
 	end
```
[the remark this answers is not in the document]
RuaridhMacd: Sorry - Thought I caught all those issues

RuaridhMacd:
It turns out that zeros(AffExpr, ...) creates an array of the same expression so any changes to one "entry" changes them all. [This is discussed here](https://jump.dev/JuMP.jl/stable/manual/expressions/#Initializing-arrays)

As such, we've gone back to @pshlime 's original loop-based method. I just moved them to a separate function to improve readability.

src/additional_tools/expression_manipulation.jl
```diff
@@ -0,0 +1,125 @@
...
+		arr[i] = AffExpr(con)
+	end
+    return nothing
+end
+
+function add_similar_to_expression!(expr1::Array{AffExpr, dims}, expr2::Array{AffExpr, dims}) where dims
+    # Check that both expr1 and expr2 have the same dimensions
+    if size(expr1) != size(expr2)
+        error("Both expressions must have the same dimensions")
+    end
```
cfe316, line 48: Please consider: 
(a) Extracting these to a function call, since it's repeated 4x
(b) (or) Having one `add_similar_to_expression!` call another, so that the 'implementation' only needs to take place once.

Also consider
* a more verbose error - would describe the sizes of the objects, and what function the error is occuring in.

src/additional_tools/expression_manipulation.jl
```diff
@@ -0,0 +1,125 @@
...
+end
+
+function createemptyexpression!(EP::Model, exprname::Symbol, dim1::Int64, dim2::Int64)
+	EP[exprname] = Matrix{AffExpr}(undef, (dim1, dim2))
+	for j=1:dim2
+		for i=1:dim1
+			EP[exprname][i,j] = AffExpr(0.0)
+		end
+	end
+end
```
cfe316, line 15: Please prefer spaces to tabs when writing new GenX functions. Indentation should be four spaces. This is to be consistent with existing GenX code.

cfe316:
Additional comment for this PR: if this is a notable change (it is, hopefully, to performance!) then please add a line to the `CHANGELOG.md` under a section

```
Changed
-------------
- Use add_to_expression! instead of the += operator for memory performance improvements (#498).
```

something like that.

RuaridhMacd:
The last requested changes are made. There are still some further opportunities to aggregate the expression manipulation functions but I'd rather push this for now and then refine it further as we learn more about how to do this.

To give a benchmark, I compared using generate_model() to build the ISONE_Trizone_FullTimeseries example system using this branch versus the main branch. The table shows the runtime and memory usage for 50 samples, measured using the @benchmark macro from BenchmarkTools.jl

The average memory usage and runtime are both better using the updated functions, though the variability is slightly increased.

|                        |              | Total time | Working time | Garbage collection time | Memory used (MB) | Allocations |
|------------------------|--------------|------------|--------------|-------------------------|------------------|-------------|
| Main branch            | Mean         | 6.543      | 4.403        | 2.140                   | 6255.3           | 87730205    |
|                        | Median       | 6.584      | 4.545        | 2.319                   | 6255.3           | 87730205    |
|                        | Standard Dev | 0.877      | 0.412        | 0.638                   | 0.0              | 0           |
| Add_to_expression      | Mean         | 6.135      | 4.252        | 1.882                   | 5610.1           | 85802832    |
|                        | Median       | 6.169      | 4.317        | 2.083                   | 5610.1           | 85802832    |
|                        | Standard Dev | 0.898      | 0.405        | 0.724                   | 0.0              | 0           |
| % Change               | Mean         | -6.2%      | -3.4%        | -12.0%                  | -10.3%           | -2.2%       |
| Negative = improvement | Median       | -6.3%      | -5.0%        | -10.2%                  | -10.3%           | -2.2%       |
|                        | Standard Dev | 2.5%       | -1.5%        | 13.5%                   | 0%               | 0%          |

src/model/generate_model.jl
```diff
@@ -88,32 +88,34 @@ function generate_model(setup::Dict,inputs::Dict,OPTIMIZER::MOI.OptimizerWithAtt
 
 	# Initialize Power Balance Expression
 	# Expression for "baseline" power balance constraint
-	@expression(EP, ePowerBalance[t=1:T, z=1:Z], 0)
+	createemptyexpression!(EP, :ePowerBalance, T, Z)
 
 	# Initialize Objective Function Expression
-	@expression(EP, eObj, 0)
+	temp = AffExpr(0.0)
+	EP[:eObj] = temp
```
cfe316, line 95: Can this be done as a single line?

RuaridhMacd: Yes, the similar code in `createemptyexpression!` uses a temporary variable to ensure type-stability but that isn't necessary here.

RuaridhMacd:
Updated the expression_manipulation.jl so that we only have to define the core loops in one place. This doesn't appear to affect performance beyond one additional function call and type-stability is maintained.

I also added some unit tests for each of the functions

cfe316:
Please move to `GenX/src/model/expression_manipulation.jl`. We can rename and put em in a `utility` folder in a small separate PR.

RuaridhMacd:
I've just made the changes we talked about on the call and that Jacob and I separately mentioned:

- Moved expression_manipulation.jl to /src/model
- Added tests to runtests.jl
- Removed some package imports in expression_manipulation_tests.jl

I also added an additional function to sum expressions but this has not been used in the rest of the code

cfe316 (approved):
Looks good; thanks for making these changes.

PR #503, opened 2023-07-31
Return path from write_outputs()

RuaridhMacd:
write_outputs() now returns the path where the results were saved after any changes by choose_output_dir(). This is useful when try to save other results to the same directory.

I also changed mkdir() to mkpath, to avoid errors were a parent directory of the intended results folder is missing.

Finally, I updated the mga outputs_path folder to be the folder created by  choose_output_dir(), not the original outputs_path


cfe316:
Hi @RuaridhMacd ,

I don't see any change to the mga outputs_path folder. I only see the first two changes you mentioned.

cfe316 (approved):
Thanks Ruaridh, this looks good. Let's squash this into `develop`.

PR #510, opened 2023-08-04
Issue templates

cfe316:
This creates templates for Bug Reports, Feature Requests, and General Questions.


cfe316:
FYI this is what it'll look like https://github.com/cfe316/GenX/issues/new/choose

PR #511, opened 2023-08-04
Feature issue template

cfe316:
This creates templates for Bug Reports, Feature Requests, and General Questions.

See https://github.com/cfe316/GenX/issues/new/choose for an example of what it will look like.


gmantegna (approved):
Looks good.

Issue #512, opened 2023-08-04
Resources in example cases may not be properly specified


cfe316:
In all of the `RealSystemExample` cases, the three small_hydroelectric_1 plants:

```
NENGREST_small_hydroelectric_1
NENG_CT_small_hydroelectric_1
NENG_ME_small_hydroelectric_1
```
have none of `[:THERM, :MUST_RUN, :STOR, :FLEX, :HYDRO, :VRE]` set: they're all zeros. I don't know how a resource would behave it has none of these set. Are these resources improperly specified? I think without any of these mutually exclusive models they would not contribute to the PowerBalance expressions.

This is true for cases
```
ISONE_SingleZone
ISONE_Trizone
ISONE_Trizone_FullTimeSeries
ISONE_Trizone_Multistage
MGA_ISONE_Trizone_FullTimeSeries
```
.

cfe316:
BTW I found this programatically using the `validate_resources` feature that I'm developing to check for these issues!

PR #513, opened 2023-08-04
Validate resources in generators_data.csv

cfe316:
Checks that the information in generators_data.csv is at least self-consistent and valid. Right now, this just checks the values in the columns THERM, LDS, FLEX, HYDRO, VRE, STOR, and LDS. It makes sure that only one of the mutually-exclusive types is present,
   and that a resource which is LDS is either STOR or HYDRO.
Also, I add the vector of resource-dictionaries (`GenXResource`s) to the inputs dictionary.

Eventually, relevant data (such as one or more time-series) could be added to each of these resource dicts.


cfe316:
Since these are written in a modular way it would be great to write tests for these, as examples for future tests.

cfe316:
We could also validate things like these:

* Zone > 0
* Resource names are unique
* Num_VRE_Bins > 0 for VRE
* Num_VRE_Bins = 0 for non-VRE
* Existing_Cap_MW, MWh, Charge_Cap_MW > 0
* Heat_Rate_MMBTU_per_MWh > 0
* Cap_Size > 0
* Up_Time, Down_Time are valid only for THERM>0
* Self_Disch > 0 valid only for STOR, HYDRO (?)
* Min_Duration / Max_Duration > 0 valid only for STOR, HYDRO ?
* LDS must be 1 or 0
* MGA must be 1 or 0

* `region` must be a string
* check that regions match with Zones

cfe316:
Note that this PR will break all of the `RealSystemExample` cases since their `small_hydro` resources have no flags set. See #512 .

gmantegna (approved):
Looks good.

Issue #427, opened 2023-04-28
Remove doubled documentation on multi-stage investment planning


cfe316:
Specifically the part here: 
https://genxproject.github.io/GenX/dev/#Multi-stage-investment-planning

dharik13:
Remove the multi-stage planning in the introduction section.

PR #515, opened 2023-08-07
Deduplicate 'multi-stage' overview docs

cfe316:
Removed from the "GenX: Introduction" page in factor of the "Multi-stage Capacity Expansion Planning with GenX" page.

This addresses #427


gmantegna (approved):
Looks good

PR #517, opened 2023-08-08
Set example small_hydro plants MUST_RUN=1

cfe316:
Turn on `MUST_RUN=1` for the `*_small_hydro` plants in the `RealSystemExample` cases. A resource flag is necessary; else the plant does not contribute to the power balance. Previously these plants had all `0`s.


RuaridhMacd (approved):
Looks good to me. All the small_hydro plants also to have availability < 1, so they're being properly limited.

PR #518, opened 2023-08-08
Remove coveralls and coverage

cfe316:
These are causing the build to fail and we don't have any tests, so they were pretty pointless. Without this we can at least see if the builds succeed.


gmantegna (approved):
Looks good

PR #520, opened 2023-08-11
Update PULL_REQUEST_TEMPLATE.md

gmantegna:
Update to fix heading formatting so all the headings look the same. This is very OCD of me but I think it makes more sense to have the headings be consistent; with the existing template it looks like the things at the end are sub-headings of the checklist.


cfe316 (approved):
Thanks, seems reasonable

PR #525, opened 2023-08-15
Feature electrolysis basic

JesseJenkins:
This feature implements a basic electrolyzer resource. 

The user either specifies a minimum quantity of hydrogen that must be produced annually by each electrolyzer and/or a price per metric tonne of hydrogen which is subtracted from the objective function. 

This implementation features hourly ramp up/down constraints and minimum and maximum hourly power consumption constraints (similar to constraints for thermal generators without unit commitment). 

It also features an hourly clean supply matching constraint that enables implementation of the 'three pillars' requirements for clean hydrogen sourcing of 'new supply/additionality', 'deliverability' and 'temporal matching.'

Demand from electrolyzers can also be added to Energy Resource Standard constraints to ensure that clean supply for electrolysis is not counted towards RPS/CES goals. 

Documentation should be complete, but please review for clarity before approval.


gmantegna (changes requested):
Looks good in terms of the constraints and test case. My one major concern is that it introduces requirements for new columns in Generators_data.csv, meaning it is not backwards compatible (also I think @cfe316 has mentioned we want to move towards reducing requirements for what is in Generators_data which makes sense). Is it possible to refactor so that this is not necessary? This could probably be accomplished similar to how retrofit resources are handled in load_generators_data. Finally please merge in the latest develop to this branch so that we can sort out any merge conflicts before merging (since this branch is branched from a relatively old commit of main).

gmantegna:
Oh also, would it be possible to add documentation on the required data inputs to data_documentation.md?

cfe316:
That's fine, we should be able to push to this branch just fine.On Aug 16, 2023 8:28 AM, Jesse Jenkins ***@***.***> wrote:
@gmantegna or @cfe316 can one of you write an appropriate entry to the CHANGELOG as well to documentation this feature addition. I'm unsure on the appropriate style for those changes.

cfe316:
I rebased this branch on the most recent `develop`. (It was an easy rebase).

cfe316 (changes requested):
This looks good. I had a few minor change requests and went ahead and did them. 

My main question now is on the modification of the ThreeZones example and whether it can be returned to being the same.

cfe316:
ELECTROLYZER is mutually exclusive with THERM, STOR, FLEX ... right? It needs to be added to the list of mutually-exclusive resources in order to not throw an error at load time.

cfe316:
@JesseJenkins by "Energy Resource Standard" did you mean "Energy Share Requirement" (`:eESR`)? Seen in a few places in comments and documentation.

gmantegna (approved):
Looks good. Thanks for addressing the changes.

cfe316 (approved):
Looks good now.

cfe316 (changes requested):
Notes
--------
* consider changing `ELECTROLYZER` column to `SIMPLE_ELECTROLYZER` or similar ("Basic" is too similar to "alkaline"....)
* fix `README.md` in the example case; right now it's the same as for a generic SmallNewEngland example

cfe316 (commented):
> -Question: we want a Three Zone example as well? I think Jacob removed the electrolyzer input tests I had provided for the New England Three Zone system to ensure it works as intended in a multi-zone case.

This is a good idea, but could we have a very-small (one day, few generators) three-zone test case? 

> HydrogenHourylMatching

Is this a typo or intentional to miss Dolphyn?

CHANGELOG.md
dharik13, line 1: Would it help to have an additional description of what the electrolyzer model does?

src/load_inputs/load_generators_data.jl
dharik13, line 1: I would suggest keeping the delete comments on start fuel cost parameter scaling since they are non-intuitive.

dharik13 (commented):
Looks good overall. Here are a few suggestions/questions:
- I would suggesting introducing a flag in the model setup to choose which type  time-matching requirement to enforce (annual vs. hourly with excess sales to start but one could easily model others like hourly without excess sales monthly which is relevant for EU for example). Under the time-matching_requirement branch (https://github.com/macroenergy/DOLPHYN/tree/time_matching_requirements) in DOLPHYN, we allow user to choose between one of  three TMRs currently: a) hourly with excess sales allowed, b) hourly with excess sales disallowed and c) annual. Something similar can be readily implemented here.
    - In electrolyzer.jl: By setting alternative values of HydrogenHourlyMatching flag (e.g. 1=Hourly, 2 = hourly with no excess sales, 3 =annual), one can enhance flexibility of the code to test multiple policy constructs.  If this is implemented, it would be more informative to change the flag label to HydrogenTimeMatching or something like that.
- A general comment: if this is meant to be a basic implementation, I would suggest adding a section in the mark up documentation for electrolyzer.jl where the modeled features are highlighted. That would also be a good place to  point them to the availability of a richer implementation in the DOLPHYN code.
- In load_generators_data.jl, I would suggest keeping the deleted comments on start fuel cost units changes with parameter scaling since they are non-intuitive (this comment assumes that this will eventually be pushed to main)
- In electrolyzer.jl:  any reason for enforcing only minimum annual demand but not daily or hourly demand? Using an annual demand might not capture the fact that the electrolyzer would often need to be ovesrized relative to avg. hourly H2 demand.

cfe316 (approved):
Looks good.

PR #528, opened 2023-08-18
Bug fix and simplify expression_manipulations.jl

RuaridhMacd:
@gmantegna identified some bugs on the develop branch after PR #498 . I had missed combinations of types in how I defined `add_similar_to_expression!()`. To fix this and try to reduce the chances of the same happening in the future, I've already:

- Abstracted the method definitions so that we don't have to explicitly define every combination. This should also hopefully reduce code maintenance requirements in the future as we use new types.
- `add_similar_to_expression()` and most of the other functions are now defined for single entry expressions as opposed to only arrays of expressions.
- `add_similar_to_expression()` is also defined for AbstractArrays, such as expressions which have indices which don't go from 1:length(array)
- `createemptyexpression()` is replaced with `create_empty_expression()` to make the naming consistent with other GenX functions.
- I'm in the process of updating the tests in expression_manipulation_tests.jl

This branch appears to work for all the SmallNewEngland and RealNewEngland examples, so others are encouraged to try it out. I'm trying to test this branch on a much wider range of problems to catch bugs. I'm also rerunning the performance comparison. I'll add a task list shortly.

-

### DIFF
--- a/src/model/expression_manipulation.jl
+++ b/src/model/expression_manipulation.jl
@@ -1,29 +1,27 @@
-function createemptyexpression!(EP::Model, exprname::Symbol, dim1::Int64)
-    temp = Vector{AffExpr}(undef, dim1)
-    for i=1:dim1::Int64
-        temp[i] = AffExpr(0.0)
-    end
-    EP[exprname] = temp
+###### ###### ###### ###### ###### ######
+# Create dense arrays of expressions filled with zeros to be added to later
+###### ###### ###### ###### ###### ######
+
+function create_empty_expression!(EP::Model, exprname::Symbol)
+    EP[exprname] = AffExpr(0.0)
+    return nothing
 end
 
-function createemptyexpression!(EP::Model, exprname::Symbol, dim1::Int64, dim2::Int64)
-    temp = Matrix{AffExpr}(undef, (dim1, dim2))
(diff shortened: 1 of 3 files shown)


src/model/expression_manipulation.jl
```diff
@@ -1,51 +1,68 @@
...
-function fill_with_zeros!(arr::Array{AffExpr, dims}) where dims
+###### ###### ###### ###### ###### ######
+# Fill dense arrays of expressions with zeros or a constant
+###### ###### ###### ###### ###### ######
+
+function fill_with_zeros!(arr::GenericAffExpr{C,T}) where {C,T}
+    arr = AffExpr(0.0)
+    return nothing
+end
+
```
cfe316, line 40: I'm not sure how this is supposed work. I tried it and it failed.

```
julia> using JuMP

julia> exp = AffExpr(1.0)
1

julia> exp
1

julia> function fill_with_zeros!(arr::GenericAffExpr{C,T}) where {C,T}
           arr = AffExpr(0.0)
           return nothing
       end
fill_with_zeros! (generic function with 1 method)

julia> fill_with_zeros!(exp)

julia> exp
1   # this should be zero
```

gmantegna (approved):
Looks good, thanks

PR #491, opened 2023-06-22
Build GenX and run example case nightly

gschivley:
Nightly test to make sure a new build of GenX can run the simple test case. Also bump checkout version and Julia version to 1.9.

PR #529, opened 2023-08-18
Build GenX and run example case nightly

gschivley:
Moving #491 to this PR. I'm not sure what branches/Julia versions are considered supported and should be included in the matrix -- @sambuddhac @cfe316 can you let me know if `main`/`develop` and 1.8/1.9 are sufficient?


gschivley:
IIRC, the cron job won't run until it is merged into `main`. Since this PR doesn't affect GenX code it might be better to switch the target branch so it can start running nightly.

sambuddhac:
I just have a slight doubt whether the target branch should be main or develop. Also, I don't think we're running with Julia 1.9 yet. But, I can set up for that.

gmantegna (changes requested):
The Test-example job is still failing on your repo. Can you look into what's going on?

gmantegna (approved):
Looks good

PR #530, opened 2023-08-21
Bug Fix for correct tracking of current investment stage by SDDP

filippopecci:
The value of `mysetup["CurrStage"]` is modified in `run_genx_case_multistage!` inside a for loop. The dictionary `mysetup` is then passed to function `run_ddp` as `setup_d` and the current stage value is used by function `initialize_cost_to_go` but it is not updated to match the current stage, and left set to the last stage as done by parent function  `run_genx_case_multistage!`. 

This commit fixes that and set the right current stage in `setup_d` before initializing the cost-to-go.


gmantegna (approved):
Looks good

Issue #514, opened 2023-08-07
"Min_Retired_Cap_MW" Behaving Badly


bpierpont:
Bug report!

In multi-stage models, "Min_Retired_Cap_MW" is getting multiplied by the number of stages. So if the parameter has a value of 100 in stage 2, the model will use a constraint of 200. In stage 5, it would turn an input of 100 into a constraint of 500.

The problem appears to be in lines 80-84 of "src/multi_stage/endogenous_retirement.jl", where it sums the current stage's value of this parameter over the number of stages (rather than summing this parameter across stages). I'm guessing lines 135 and 184 of that file have similar issues but I haven't tested it.

cfe316:
This is now fixed on `develop`.

PR #542, opened 2023-09-05
Added computation of cumulative minimum capacity retirements in multistage

filippopecci:
Fixed computation of cumulative minimum retirements in multistage code. Previously, minimum capacity retirement parameters were not added up and this resulted in the minimum retirement constraints enforcing wrong lower bounds ( #514 )

### DIFF
--- a/src/multi_stage/endogenous_retirement.jl
+++ b/src/multi_stage/endogenous_retirement.jl
@@ -22,6 +22,47 @@ function get_retirement_stage(cur_stage::Int, lifetime::Int, stage_lens::Array{I
     return Int(ret_stage)
 end
 
+function update_cumulative_min_ret!(inputs_d::Dict,t::Int,Resource_Set::String,dfGen_Name::String,RetCap::Symbol)
+
+	CumRetCap = Symbol("Cum_"*String(RetCap));
+
+	if !isempty(inputs_d[1][Resource_Set])
+		if t==1
+			inputs_d[t][dfGen_Name][!,CumRetCap] = inputs_d[t][dfGen_Name][!,RetCap];
+		else
+			inputs_d[t][dfGen_Name][!,CumRetCap] = inputs_d[t-1][dfGen_Name][!,CumRetCap] + inputs_d[t][dfGen_Name][!,RetCap];
+		end
+	end
+
+end
+
+
+function compute_cumulative_min_retirements!(inputs_d::Dict,t::Int)
+
+	mytab =[("G","dfGen",:Min_Retired_Cap_MW),
+	("STOR_ALL","dfGen",:Min_Retired_Energy_Cap_MW),
+	("STOR_ASYMMETRIC","dfGen",:Min_Retired_Charge_Cap_MW)];
+
+	if !isempty(inputs_d[1]["VRE_STOR"])
+		append!(mytab,[("VS_DC","dfVRE_STOR",:Min_Retired_Cap_Inverter_MW),
+				("VS_SOLAR","dfVRE_STOR",:Min_Retired_Cap_Solar_MW),
+				("VS_WIND","dfVRE_STOR",:Min_Retired_Cap_Wind_MW),
+				("VS_STOR","dfGen",:Min_Retired_Energy_Cap_MW),
+				("VS_ASYM_DC_DISCHARGE","dfVRE_STOR",:Min_Retired_Cap_Discharge_DC_MW),
+				("VS_ASYM_DC_CHARGE","dfVRE_STOR",:Min_Retired_Cap_Charge_DC_MW),
+				("VS_ASYM_AC_DISCHARGE","dfVRE_STOR",:Min_Retired_Cap_Discharge_AC_MW),
+				("VS_ASYM_AC_CHARGE","dfVRE_STOR",:Min_Retired_Cap_Charge_AC_MW)])
+
+	end
+
+	for (Resource_Set,dfGen_Name,RetCap) in mytab
+		update_cumulative_min_ret!(inputs_d,t,Resource_Set,dfGen_Name,RetCap)
+	end
+
+		
+end
+
+
 function endogenous_retirement!(EP::Model, inputs::Dict, setup::Dict)
 	multi_stage_settings = setup["MultiStageSettingsDict"]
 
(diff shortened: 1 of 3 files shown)


src/multi_stage/endogenous_retirement.jl
```diff
@@ -22,6 +22,56 @@ function get_retirement_stage(cur_stage::Int, lifetime::Int, stage_lens::Array{I
     return Int(ret_stage)
 end
 
+function compute_min_cumulative_retirements!(t::Int,inputs_d::Dict)
```
cfe316, line 25: To make the code easier to read, consider adding a simple one-line function like

`dfgen(stage) = inputs_d[stage]["dfGen"]`

and also 

```
dfvrestore(stage) = inputs_d[stage]["dfVRE_STOR"]
```

This also has the advantage that it makes the interface smaller. If you ever want to change `"dfVRE_STOR"` to another key it's easier since it's only in one place.

cfe316: Could also do something like

```
function xxx(a,b,c):
    if !isempty(a)
        inputs[...][!, Symbol(b)] = sum( .... Symbol(c) for p in 1:t)
     end
     
table = [("VS_DC", "Min_Cum_Retired_Cap_Inverter_MW", "Min_Retired_Cap_Inverter_MW"),
              (..., ..., ...),
              ]
for row in table:
    a,b,c = row
    xxx(a,b,c)
end
```

lbonaldo: Two more suggestions to improve readability in addition to @cfe316 comments:

- You could use a `Symbol` to select a column (e.g.,`:Min_Cum_Retired_Energy_Cap_MW`) instead of converting a `String` into a `Symbol` every time, (e.g., `Symbol("Min_Cum_Retired_Energy_Cap_MW")`)

- The usual convention for mutating functions (e.g., `myfun!(t::Int,inputs_d::Dict)` is that the argument that is modified is placed first (Ref.: https://docs.julialang.org/en/v1/manual/style-guide/#Write-functions-with-argument-ordering-similar-to-Julia-Base)
```suggestion
function compute_min_cumulative_retirements!(inputs_d::Dict, t::Int)
```

src/multi_stage/endogenous_retirement.jl
```diff
@@ -22,6 +22,56 @@ function get_retirement_stage(cur_stage::Int, lifetime::Int, stage_lens::Array{I
...
+
+	inputs_d[t]["dfGen"][!,Symbol("Min_Cum_Retired_Cap_MW")] = sum(inputs_d[p]["dfGen"][!,Symbol("Min_Retired_Cap_MW")] for p=1:t);
```
lbonaldo, line 27: To improve performance, If `t` can be very large, we could think about a way to avoid redoing the sum over all the keys `1:t` of the dictionary, since the sum over `1:t-1` is stored at `inputs_d[t-1]["dfGen"][!,:Min_Cum_Retired_Cap_MW]`. Am I correct?

src/multi_stage/endogenous_retirement.jl
```diff
@@ -22,6 +22,56 @@ function get_retirement_stage(cur_stage::Int, lifetime::Int, stage_lens::Array{I
...
+	if !isempty(inputs_d[1]["STOR_ALL"])
+		inputs_d[t]["dfGen"][!,Symbol("Min_Cum_Retired_Energy_Cap_MW")] = sum(inputs_d[p]["dfGen"][!,Symbol("Min_Retired_Energy_Cap_MW")] for p=1:t)
+	end
+
+	if !isempty(inputs_d[1]["STOR_ASYMMETRIC"])
+		inputs_d[t]["dfGen"][!,Symbol("Min_Cum_Retired_Charge_Cap_MW")] = sum(inputs_d[p]["dfGen"][!,Symbol("Min_Retired_Charge_Cap_MW")] for p=1:t)
+	end
+
+	if !isempty(inputs_d[1]["VRE_STOR"])
+		if !isempty(inputs["VS_DC"])
```
lbonaldo, line 38: Is this a typo? 

```suggestion
		if !isempty(inputs_d[1]["VS_DC"])
```

filippopecci:
Thank you @cfe316 and @lbonaldo for your suggestions. I have implemented all of them, have a look and let me know.

src/multi_stage/endogenous_retirement.jl
```diff
@@ -22,6 +22,51 @@ function get_retirement_stage(cur_stage::Int, lifetime::Int, stage_lens::Array{I
...
+				("VS_DC","dfVRE_STOR",:Min_Retired_Cap_Inverter_MW),
+				("VS_SOLAR","dfVRE_STOR",:Min_Retired_Cap_Solar_MW),
+				("VS_WIND","dfVRE_STOR",:Min_Retired_Cap_Wind_MW),
+				("VS_STOR","dfGen",:Min_Retired_Energy_Cap_MW),
+				("VS_ASYM_DC_DISCHARGE","dfVRE_STOR",:Min_Retired_Cap_Discharge_DC_MW),
+				("VS_ASYM_DC_CHARGE","dfVRE_STOR",:Min_Retired_Cap_Charge_DC_MW),
+				("VS_ASYM_AC_DISCHARGE","dfVRE_STOR",:Min_Retired_Cap_Discharge_AC_MW),
+				("VS_ASYM_AC_CHARGE","dfVRE_STOR",:Min_Retired_Cap_Charge_AC_MW)
+			]
+	end
```
cfe316, line 60: Is the first `mytab` a subset of the second one? Could this be converted to 
```
mytab = ...

if VRESTORE
   add new entries to my_tab
end
```

Then they'd only need to be listed once.

src/multi_stage/endogenous_retirement.jl
```diff
@@ -22,6 +22,51 @@ function get_retirement_stage(cur_stage::Int, lifetime::Int, stage_lens::Array{I
...
+function update_cumulative_min_ret!(inputs_d::Dict,t::Int,Resource_Set::String,dfGen_Name::String,RetCap::Symbol)
+
+	CumRetCap = Symbol("Cum_"*String(RetCap));
+
+	if !isempty(inputs_d[1][Resource_Set])
+		if t==1
+			inputs_d[t][dfGen_Name][!,CumRetCap] = inputs_d[t][dfGen_Name][!,RetCap];
+		else
+			inputs_d[t][dfGen_Name][!,CumRetCap] = inputs_d[t-1][dfGen_Name][!,CumRetCap] + inputs_d[t][dfGen_Name][!,RetCap];
+		end
```
cfe316, line 34: It looks like the second line here is the sum of the first line plus the term `inputs_d[t-1][dfGen_Name][!,CumRetCap]`

Could this be reformulated to 
```
A = B
if t==1
   A += C
end
```

cfe316 (approved):
Thanks for the fix!

PR #545, opened 2023-09-07
Remove unneeded check in `load fuels`

cfe316:
minor cleanup


gmantegna (commented):
Why is the for loop still needed if it is looping over just one thing?

cfe316:
In case we needed to add a lowercase `none` as well or something similar.

gmantegna (approved):
good job !

PR #546, opened 2023-09-09
Thermal+storage (fusion, maintenance)

cfe316:
This is a draft pull request, just to show everyone what might should be coming down the pipeline soon. The branch is currently based on v0.3.6, not on develop, so that why I've done it this way for now.

I _am_ pretty happy with how I wrote up the formulation `thermal_storage.jl`. There are some new techniques here.
I'm also happy with the more simplified output files as seen in `write_thermal_storage.jl`.

TBD:
* A serious capacity_reserve_margin+storage formulation.
* Maybe change maintenance to act on the generator rather than the thermal core?
* Or perhaps even allow a separate formulation for each? (they'll have to sync up, optimally!)
* Documentation
* Existing capacities and retirements, not just new-build.


cfe316:
Closing because this is not ready, and will need to be substantially reworked.

PR #551, opened 2023-09-19
Change ModelParameterScale to const

cfe316:
"const is used to declare global variables whose values will not change. In almost all code (and particularly performance sensitive code) global variables should be declared constant in this way." 

https://docs.julialang.org/en/v1/base/base/#const


lbonaldo (approved):
Great. Also a useful reading: https://docs.julialang.org/en/v1/manual/performance-tips/#Avoid-untyped-global-variables

PR #577, opened 2023-11-12
Zero-out Rsv_Max and Reg_Max of MUST_RUN example gens

cfe316:
This prevents them from contributing an unbounded amount of vREG and vRSV when Reserves are on.


lbonaldo (approved):
Thanks @cfe316 for fixing this!

PR #578, opened 2023-11-13
Remove MoM test

lbonaldo:
This removes the test for the MoM. The test will be reinstated for an updated version of the module.


cfe316 (approved):
Thanks for removing this. Unfortunately due to the random nature of MoM we need to find a better way to test this in the future.

PR #579, opened 2023-11-14
Extend expression_manipulations.jl to some variables

RuaridhMacd:
As pointed out in PR #562  and elsewhere, there are still some issues with the expression_manipulations functions. This PR tries to resolve the following:

1. There are some undefined functions, e.g. _sum_expressions()
2. There are some unnecessary duplicate methods with concrete instead of abstract types 
3. Some bugs exist around edge-cases with various combinations of types
4. Some of the methods don't work with variables when it seems they should, e.g. add_similar_to_expression()
5. The methods aren't documented.
6. The sum_expression() methods don't have tests

@cfe316 has addressed issue number 4 in PR #562  but I also included a quick way of handling variables using the `1 * ` dummy coefficient that he proposed.

There are still some extensions we could make, in particular creating arrays using generic expressions and variables which don't use Float64 coefficients and constants or that have non-dense indices.

The SmallNewEngland cases worked for me with these changes but I'd definitely appreciate help checking for any edge cases.


RuaridhMacd:
One note: I simplified some of the indexing using `eachindex()`. This sacrifices a small amount of speed / memory usage but makes the methods work for AbstractArrays with sparse indices and reduces the number of functions we need to maintain.

RuaridhMacd:
Converting this to a draft as there's an issue with the methods which take GenericVariableRef. The methods work but then don't save the new expression in place of the variable. I'll look at a fix.

RuaridhMacd:
I've removed the functions which worked on variables, rather than expressions. They don't work as in-place functions. We can perhaps add new functions in the future to help users go from variables -> expressions

RuaridhMacd:
@lbonaldo @sambuddhac I think this should be ok for v0.4

src/model/expression_manipulation.jl
```diff
@@ -33,15 +44,28 @@ end
 # Fill dense arrays of expressions with zeros or a constant
 ###### ###### ###### ###### ###### ######
 
-function fill_with_zeros!(arr::Array{GenericAffExpr{C,T}, dims}) where {C,T,dims}
+@doc raw"""
+    fill_with_zeros!(arr::Array{GenericAffExpr{C,T}, dims}) where {C,T,dims}
```
lbonaldo, line 48:
```suggestion
    fill_with_zeros!(arr::AbstractArray{GenericAffExpr{C,T}, dims}) where {C,T,dims}
```

src/model/expression_manipulation.jl
```diff
@@ -33,15 +44,28 @@ end
...
     for i::Int64 in eachindex(IndexLinear(), arr)::Base.OneTo{Int64}
         arr[i] = AffExpr(0.0)
     end
     return nothing
 end
 
-function fill_with_const!(arr::Array{GenericAffExpr{C,T}, dims}, con::Real) where {C,T,dims}
-    for i::Int64 in eachindex(IndexLinear(), arr)::Base.OneTo{Int64}
+@doc raw"""
+    fill_with_const!(arr::Array{GenericAffExpr{C,T}, dims}) where {C,T,dims}
```
lbonaldo, line 60:
```suggestion
    fill_with_const!(arr::AbstractArray{GenericAffExpr{C,T}, dims}, con::Real) where {C,T,dims}
```

lbonaldo (approved):
Thanks @RuaridhMacd for this PR. I just added a couple of small comments.

PR #580, opened 2023-11-15
Separate reserves into creation and constraints

cfe316:
Divide the core Reserves functionality into

1. variable and expression creation

(space for potential modification)

2. constraint application

This makes the module more like the CapacityReserveMargin or MinCapReq policies, allowing the expressions to be modified before constraints are applied.


cfe316:
Note that this PR does not interact at all with #562.

lbonaldo:
Thanks for this change Jacob!

lbonaldo (approved):

PR #582, opened 2023-11-22
Function to compute locational marginal prices from output

cfe316:
This slightly simplifies
`write_prices.jl`
`write_charging_cost.jl`
`write_energy_revenue.jl`.

I tested this on SNE/ThreeZones and the results are identical to within numerical errors.


lbonaldo (approved):
This looks good. Thanks @cfe316!

PR #591, opened 2023-12-03
Prefer columns-style network format to matrix format

cfe316:
This changes the preferred Network input format from a matrix of 1, 0, -1 to lists of Start_Node, End_Node.

* This may not be perfect: here we use Node, which a more technical term, but in other places in the code and docs we largely use "Zone".


cfe316:
I've changed back to using Zone.

Issue #485, opened 2023-06-17
write_net_revenue.jl neglects investment and FOM costs for ASYMMETRIC storage.


AshwinSandeep1:
The cost summation is missing 
1. Charging Investment cost
2. Charging FOM cost

Issue #571, opened 2023-11-05
Investment and fixed charge costs missing from `NetRevenue.csv` output


ASamarkRoth:
Should add line:

`dfNetRevenue.Inv_cost_Charge_MW = dfGen[!,:Inv_Cost_Charge_per_MWyr] .* dfCap[1:end-1,:NewChargeCap]` 

after line 19 and:

`dfNetRevenue.Fixed_OM_cost_Charge_MW = dfGen[!,:Fixed_OM_Cost_Charge_per_MWyr] .* dfCap[1:end-1,:EndChargeCap]`

after line 27 in https://github.com/GenXProject/GenX/blob/main/src/write_outputs/write_net_revenue.jl 

as well as incorporating these additions to the costs calculation at the end of the function: 

`dfNetRevenue.Cost = dfNetRevenue.Inv_cost_MW .+ dfNetRevenue.Inv_cost_MWh .+ dfNetRevenue.Inv_cost_Charge_MW .+ dfNetRevenue.Fixed_OM_cost_MW .+ dfNetRevenue.Fixed_OM_cost_MWh .+ dfNetRevenue.Fixed_OM_cost_Charge_MW .+ dfNetRevenue.Var_OM_cost_out .+ dfNetRevenue.Var_OM_cost_in .+ dfNetRevenue.Fuel_cost .+ dfNetRevenue.Charge_cost .+ dfNetRevenue.EmissionsCost .+ dfNetRevenue.StartCost`

cfe316:
Thanks Anton. This is the same issue as #485.

sambuddhac:
This is already taken care of, in the `develop` branch. Will be reflected on `main` after the merge of `develop` to `main` (followed by release of version v0.4)

PR #602, opened 2023-12-13
Fix-485: Account for Inv_Cost_Charge and Fixed_OM_Cost_Charge for STOR=2, in NetRevenue.csv

cfe316:
This fixes the issues #485 and #571 .


cfe316:
I've tried this on a modified SmallNewEngland/OneZone case. I expected the Profit column of the asymmetric storage to sum to zero, but it came out negative. I might like someone else to check this.

lbonaldo (approved):

cfe316:
In RealSystemExample/ISONE_Trizone, some storage has negative profit. This seems to be normal enough.

Issue #605, opened 2024-01-06
Possible bug when using exogenous capacities in multi-stage models with separation of new build and can retire.


filippopecci:
PR[#326](https://github.com/GenXProject/GenX/pull/326) has allowed the user to fix generator capacity exogenously in multi-stage models, by tracking capacity between stages only for generators in RET_CAP, which included both bildable and retirable generators. The changes introduced by PR[#392](https://github.com/GenXProject/GenX/pull/392) separate NEW_BUILD and CAN_RETIRE, allowing a generator to be buildable but not retirable.

As a result, files `dual_dynamic_programming.jl` and `endogenous_retirement.jl` should be checked and modified accordingly. Generally speaking, we could substitute RET_CAP in those functions with a set VAR_CAP = union(RET_CAP,NEW_CAP).

filippopecci:
This issue was fixed by PR #609

PR #609, opened 2024-01-08
Fixed issue #605

filippopecci:
The multistage algorithm will track capacity between stages for all generators with either New_Build=1 or Can_Retire=1.


lbonaldo (approved):
Thank you, @filippopecci!

PR #610, opened 2024-01-09
Crunching the mass-based and demand-rate based into one expression in…

sambuddhac:
… write_net_revenue.jl

The expressions are exactly the same


lbonaldo (approved):
Thanks, @sambuddhac. This looks good to me. You could also remove the commented lines unless they are still meaningful.

PR #615, opened 2024-01-23
Update Julia versions in ci.yml

lbonaldo:
This updates the list of Julia versions in the ci.yml file and removes the tests on Julia nightly, as they are not allowed to fail. Reference: https://github.com/actions/runner/issues/2347

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,16 +17,13 @@ jobs:
           - "1.6" # LTS (64-bit Linux)
           - "1.7"
           - "1.8" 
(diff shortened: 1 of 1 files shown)


sambuddhac (approved):
Okay, I am now convinced of its value. Approving it !!!

sambuddhac:
Okay, I am now convinced of its value. Approving it !!!

lbonaldo:
Thank you, @sambuddhac! In the future, we might add back the tests on the Julia nightly version if the `allow_failure: true` gets implemented or as a separate test.

Issue #553, opened 2023-09-22
Weird behavior in cost writing when objective is changed


ml6802:
When running MGA, cost files are not reliable for several reasons, some seem to be intentional and should be upgraded, while others seem buggy.

- cTotal is based on the value of the objective function, not a separate expression. Making it reliably a separate expression will allow for accurate total cost quantification when running MOO or MGA methods.

- When MGA is run, other cost numbers look funky too for unclear reasons: examples are:

1. Sometimes the variable costs come out negative - unclear why

2. the sum of zonal costs for non cost-optimal runs are occasionally less than those for the cost optimal run. This might be due to system costs not caught in the sum of zonal costs.

lbonaldo:
I created a PR targeting **develop** to address the first of the two fixes above. @ml6802, could you please test the change with the cases you are running and see if #620 fixes `cTotal`? If yes, then this bug is not present in **main**.

sambuddhac:
This is now fixed on `develop`

PR #620, opened 2024-02-01
Fix cTotal on write_costs: #553

lbonaldo:
This PR addresses the first part of issue #553.
The "Total" column of the "cTotal" row of `costs.csv` is now equal to `value(EP[:eObj])` instead of the model objective function. This ensures that `costs.csv` contains the total cost even when the objective function changes (e.g., in MGA).


sambuddhac (approved):
reviewed and approved !!!

Issue #616, opened 2024-01-29
Tx losses file has value larger than flows.csv


gschivley:
The transmission losses reported from my model are a couple orders of magnitude larger than the absolute flows for each line. It looks like [this commit](https://github.com/GenXProject/GenX/commit/007c7271c54e3e8e445d65f20fd8b57d82d61016) changed the output calculation from multiplying by `ModelScalingFactor` to multiplying by `ModelScalingFactor^2`. 

@cfe316, am I reading the code correctly?

cfe316:
While I believe your statement that there are anomalously high losses, the output even before that commit multiplied by ModelScalingFactor twice!

 ```
 for i in 1:L
	if i in inputs["LOSS_LINES"]
		tlosses[i,:] = value.(EP[:vTLOSS])[i,:] * ModelScalingFactor     # once
	end
	...
end
dfTLosses = hcat(dfTLosses, DataFrame(tlosses * ModelScalingFactor, :auto))     # twice
```

lbonaldo:
Thank you, @gschivley, for opening this issue! Fixed in #622 (main) and #621 (develop).

PR #622, opened 2024-02-02
Fix scaling of tx losses in tlosses.csv #616

lbonaldo:
This PR fixes the scaling of the transmission losses due to power flows in write_transmission_losses.jl (#616)


sambuddhac (approved):
Seems good to me !!!

PR #624, opened 2024-02-07
Compute conflicting constraints for infeasible models

filippopecci:
Implemented functions to compute conflicting constraints when model is infeasible


lbonaldo (approved):
Thanks, @filippopecci, for adding this feature!

PR #626, opened 2024-02-13
Fix write_net_revenue when Reserves = 0

lbonaldo:
This fixes write_net_revenue.jl and write_outputs.jl when Reserves = 0


sambuddhac (approved):
Ah !!! Sorry to have missed that nudge. Thanks for fixing. Great catch !!!

PR #654, opened 2024-03-18
Added docstring for endogenous retirement constraint

sambuddhac:
Added the constraint equation

CAUTION: Need to make sure about the notation


lbonaldo (approved):
Thanks @sambuddhac!

PR #657, opened 2024-03-19
Obj scale factor

Betristor:
Sometimes the model will be built into an ill form if some objective terms are quite large or small. To alleviate this problem, we could add a scaling factor to scale the objective function during solving while leaving all other expressions untouched. In SmallNewEngland/OneZone_3VREBin case, an example ObjScale of 10 could help to change the objective range from [1e-4, 5e+02] to [1e-3, 5e+03].


Betristor:
@sambuddhac Thanks, I encountered numerical issues during the solving process several times, and thus I hope this fix can help others avoid these numerical issues conveniently and efficiently.

sambuddhac:
Dear @Betristor , now is a good time to see if you still have the problem and if yes, accordingly revise your PR. We just did a prerelease of GenX v0.4.0. I would suggest you to do a fresh clone and run from the main. Please observe that this new version comes with a complete redressing of the input  data. So, you would have to modify your test case. Here's a translator package from old to new data format: https://github.com/lbonaldo/GenX/tree/release/0.4.0_user_guide/split_generators_data
Please let me know if you need further help.

Betristor:
@sambuddhac Hi, Sam, I updated my code with the latest development, but still it introduces too many changes. I shall close this PR, turn to a new branch and commit the changes.

PR #773, opened 2024-09-23
[DO NOT MERGE] various experiments with performance benchmarking

odow:
I'm discussing some things with @lbonaldo 

This shouldn't be merged, but we might find some things that could be extracted into new PRs.

### DIFF
--- a/src/model/core/co2.jl
+++ b/src/model/core/co2.jl
@@ -74,12 +74,13 @@ function co2!(EP::Model, inputs::Dict)
     if isempty(CCS)
         @expression(EP, eEmissionsByPlant[y = 1:G, t = 1:T],
             if y in SINGLE_FUEL
-                ((1 - biomass(gen[y])) * (EP[:vFuel][y, t] + EP[:vStartFuel][y, t]) *
-                 fuel_CO2[fuel(gen[y])])
+                (1 - biomass(gen[y])) * fuel_CO2[fuel(gen[y])] * (EP[:vFuel][y, t] + EP[:vStartFuel][y, t])
             else
-                sum(((1 - biomass(gen[y])) *
-                     (EP[:vMulFuels][y, i, t] + EP[:vMulStartFuels][y, i, t]) *
-                     fuel_CO2[fuel_cols(gen[y], tag = i)]) for i in 1:max_fuels)
+                sum(
+                    (1 - biomass(gen[y])) * fuel_CO2[fuel_cols(gen[y], tag = i)] *
+                    (EP[:vMulFuels][y, i, t] + EP[:vMulStartFuels][y, i, t])
+                    for i in 1:max_fuels
+                )
             end)
     else
         @info "Using the CO2 module to determine the CO2 emissions of CCS-equipped plants"
--- a/src/model/core/co2.jl
+++ b/src/model/core/co2.jl
@@ -1,46 +1,46 @@
-@doc raw""" 
+@doc raw"""
     co2!(EP::Model, inputs::Dict)
 
-This function creates expressions to account for CO2 emissions as well as captured and sequestrated 
-CO2 from thermal generators. It also has the capability to model the negative CO2 emissions 
-from bioenergy with carbon capture and storage. 
+This function creates expressions to account for CO2 emissions as well as captured and sequestrated
+CO2 from thermal generators. It also has the capability to model the negative CO2 emissions
+from bioenergy with carbon capture and storage.
 
 ***** Expressions *****
 
-For thermal generators which combust fuels (e.g., coal, natural gas, and biomass), the net CO2 
-emission to the environment is a function of fuel consumption, CO2 emission factor, CO2 capture 
-fraction, and whether the feedstock is biomass. Biomass is a factor in this equation because 
-biomass generators are assumed to generate zero net CO2 emissions, or negative net CO2 emissions 
+For thermal generators which combust fuels (e.g., coal, natural gas, and biomass), the net CO2
+emission to the environment is a function of fuel consumption, CO2 emission factor, CO2 capture
+fraction, and whether the feedstock is biomass. Biomass is a factor in this equation because
+biomass generators are assumed to generate zero net CO2 emissions, or negative net CO2 emissions
 in the case that the CO2 they emit is captured and sequestered underground.
 
 If a user wishes to represent a generator that combusts biomass, then in the resource .csv files,
 the "Biomass" column (boolean, 1 or 0), which represents if a generator $y$ uses biomass or not, should be set to 1.
 The CO2 emissions from such a generator will be assumed to be zero without CCS and negative with CCS.
 
-The CO2 emissions from generator $y$ at time $t$ are determined by total fuel 
-consumption (MMBTU) multiplied by the CO2 content of the fuel (tCO2/MMBTU), and by 
-(1 - Biomass [0 or 1] - CO2 capture fraction [a fraction, between 0 - 1]). 
+The CO2 emissions from generator $y$ at time $t$ are determined by total fuel
+consumption (MMBTU) multiplied by the CO2 content of the fuel (tCO2/MMBTU), and by
+(1 - Biomass [0 or 1] - CO2 capture fraction [a fraction, between 0 - 1]).
 The CO2 capture fraction could be differernt during the steady-state and startup events
 (generally startup events have a lower CO2 capture fraction), so we use distinct CO2 capture fractions
-to determine the emissions. 
+to determine the emissions.
 In short, the CO2 emissions for a generator depend on the CO2 emission factor from fuel combustion,
 the CO2 capture fraction, and whether the generator uses biomass.
 
 ```math
 \begin{aligned}
-eEmissionsByPlant_{g,t} = (1-Biomass_y-  CO2\_Capture\_Fraction_y) * vFuel_{y,t}  * CO2_{content} + (1-Biomass_y-  CO2\_Capture\_Fraction\_Startup_y) * eStartFuel_{y,t} * CO2_{content} 
+eEmissionsByPlant_{g,t} = (1-Biomass_y-  CO2\_Capture\_Fraction_y) * vFuel_{y,t}  * CO2_{content} + (1-Biomass_y-  CO2\_Capture\_Fraction\_Startup_y) * eStartFuel_{y,t} * CO2_{content}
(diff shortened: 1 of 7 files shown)


odow:
One issue is https://github.com/jump-dev/MutableArithmetics.jl/issues/302

It doesn't explain the 1.9/1.10 issue.

But I wonder if that is due to some interaction with the GC. A lot of the time was in `intersect` with `Vector{String}`.

src/model/core/co2.jl
```diff
@@ -134,7 +134,10 @@ function co2!(EP::Model, inputs::Dict)
     end
 
     # emissions by zone
+    RESOURCES_BY_ZONE = map(1:Z) do z
+        return resources_in_zone_by_rid(gen, z)
+    end
     @expression(EP, eEmissionsByZone[z = 1:Z, t = 1:T],
-        sum(eEmissionsByPlant[y, t] for y in resources_in_zone_by_rid(gen, z)))
+        sum(eEmissionsByPlant[y, t] for y in RESOURCES_BY_ZONE[z]))
```
odow, line 141: `resources_in_zone_by_rid` is:
```julia
function resources_in_zone_by_rid(rs::Vector{<:AbstractResource}, zone::Int)
    return resource_id.(rs[zone_id.(rs) .== zone])
end
```
This was being called for every `z` (needed) and every `t` (not needed). When `T` is large, this is expensive.

src/model/core/discharge/investment_discharge.jl
```diff
@@ -68,34 +68,39 @@ function investment_discharge!(EP::Model, inputs::Dict, setup::Dict)
...
     end
 
+    SET_1 = intersect(NEW_CAP, RET_CAP, RETROFIT_CAP) # Resources eligible for new capacity, retirements and being retrofitted
+    SET_2 = intersect(setdiff(RET_CAP, NEW_CAP), setdiff(RET_CAP, RETROFIT_CAP)) # Resources eligible for only capacity retirements
+    SET_3 = setdiff(intersect(RET_CAP, NEW_CAP), RETROFIT_CAP) # Resources eligible for retirement and new capacity
+    SET_4 = setdiff(intersect(RET_CAP, RETROFIT_CAP), NEW_CAP) # Resources eligible for retirement and retrofitting
+    SET_5 = intersect(setdiff(NEW_CAP, RET_CAP), setdiff(NEW_CAP, RETROFIT_CAP))  # Resources eligible for only new capacity
     @expression(EP, eTotalCap[y in 1:G],
-        if y in intersect(NEW_CAP, RET_CAP, RETROFIT_CAP) # Resources eligible for new capacity, retirements and being retrofitted
+        if y in SET_1
```
odow, line 77: This intersection was being re-computed for every value of `y in 1:G`. Ditto with the other ones.

src/model/core/fuel.jl
```diff
@@ -225,16 +225,19 @@ function fuel!(EP::Model, inputs::Dict, setup::Dict)
...
                 MULTI_FUELS)))
     end
 
+    RESOURCES_BY_FUEL = map(1:NUM_FUEL) do f
+        return intersect(resources_with_fuel(gen, fuels[f]), SINGLE_FUEL)
+    end
     @expression(EP, eFuelConsumption_single[f in 1:NUM_FUEL, t in 1:T],
         sum(EP[:vFuel][y, t] + EP[:eStartFuel][y, t]
-        for y in intersect(resources_with_fuel(gen, fuels[f]), SINGLE_FUEL)))
+        for y in RESOURCES_BY_FUEL[f]))
```
odow, line 240: This intersection was being re-computed for every value of `t`

src/model/resources/curtailable_variable_renewable/curtailable_variable_renewable.jl
```diff
@@ -33,13 +33,15 @@ function curtailable_variable_renewable!(EP::Model, inputs::Dict, setup::Dict)
...
     ### Expressions ###
 
     ## Power Balance Expressions ##
 
-    @expression(EP, ePowerBalanceDisp[t = 1:T, z = 1:Z],
-        sum(EP[:vP][y, t] for y in intersect(VRE, resources_in_zone_by_rid(gen, z))))
-    add_similar_to_expression!(EP[:ePowerBalance], EP[:ePowerBalanceDisp])
+    @expression(EP, eGenerationByVRE[t = 1:T, z = 1:Z],
+        sum(EP[:vP][y, t] for y in VRE_BY_ZONE[z]))
+    add_similar_to_expression!(EP[:ePowerBalance], EP[:eGenerationByVRE])
```
odow, line 44: The intersection was being re-computed for every value of `t`.

src/model/resources/curtailable_variable_renewable/curtailable_variable_renewable.jl
```diff
@@ -77,11 +79,10 @@ function curtailable_variable_renewable!(EP::Model, inputs::Dict, setup::Dict)
...
     end
-    ##CO2 Polcy Module VRE Generation by zone
-    @expression(EP, eGenerationByVRE[z = 1:Z, t = 1:T], # the unit is GW
-        sum(EP[:vP][y, t]
-        for y in intersect(inputs["VRE"], resources_in_zone_by_rid(gen, z))))
-    add_similar_to_expression!(EP[:eGenerationByZone], eGenerationByVRE)
+    ##CO2 Policy Module VRE Generation by zone
+    # We use the transpose here because eGenerationByZone is [1:Z, 1:T] and
+    # eGenerationByVRE is [1:T, 1:Z].
+    add_similar_to_expression!(EP[:eGenerationByZone], eGenerationByVRE')
```
odow, line 85: Unless I've missed something, `eGenerationByVRE` is just the transpose of `ePowerBalanceDisp` above?

src/model/resources/thermal/thermal.jl
```diff
@@ -20,10 +20,13 @@ function thermal!(EP::Model, inputs::Dict, setup::Dict)
...
         thermal_no_commit!(EP, inputs, setup)
     end
+    THERM_ALL_BY_ZONE = map(1:Z) do z
+        return intersect(inputs["THERM_ALL"], resources_in_zone_by_rid(gen, z))
+    end
     ##CO2 Polcy Module Thermal Generation by zone
     @expression(EP, eGenerationByThermAll[z = 1:Z, t = 1:T], # the unit is GW
         sum(EP[:vP][y, t]
-        for y in intersect(inputs["THERM_ALL"], resources_in_zone_by_rid(gen, z))))
+        for y in THERM_ALL_BY_ZONE[z]))
```
odow, line 29: This intersection was being recomputed for every value of `t`

src/model/resources/thermal/thermal_no_commit.jl
```diff
@@ -52,13 +52,15 @@ function thermal_no_commit!(EP::Model, inputs::Dict, setup::Dict)
...
+    THERM_NO_COMMIT_BY_ZONE = map(1:Z) do z
+        return intersect(THERM_NO_COMMIT, resources_in_zone_by_rid(gen, z))
+    end
     ### Expressions ###
 
     ## Power Balance Expressions ##
     @expression(EP, ePowerBalanceThermNoCommit[t = 1:T, z = 1:Z],
         sum(EP[:vP][y, t]
-        for y in intersect(THERM_NO_COMMIT, resources_in_zone_by_rid(gen, z))))
+        for y in THERM_NO_COMMIT_BY_ZONE[z]))
```
odow, line 63: This intersection was being recomputed for every value of `t`

src/model/core/co2.jl
```diff
@@ -69,21 +69,22 @@ function co2!(EP::Model, inputs::Dict)
...
     # CO2 emissions from power plants in "Generators_data.csv"
-    # If all the CO2 capture fractions from Generators_data are zeros, the CO2 emissions from thermal generators are determined by fuel consumption times CO2 content per MMBTU 
+    # If all the CO2 capture fractions from Generators_data are zeros, the CO2 emissions from thermal generators are determined by fuel consumption times CO2 content per MMBTU
 
     if isempty(CCS)
         @expression(EP, eEmissionsByPlant[y = 1:G, t = 1:T],
             if y in SINGLE_FUEL
-                ((1 - biomass(gen[y])) * (EP[:vFuel][y, t] + EP[:vStartFuel][y, t]) *
-                 fuel_CO2[fuel(gen[y])])
+                (1 - biomass(gen[y])) * fuel_CO2[fuel(gen[y])] * (EP[:vFuel][y, t] + EP[:vStartFuel][y, t])
```
odow, line 77: This is a weird thing that probably doesn't make a difference, but it's worth explaining.

JuMP processes from left to right.

You have `a::Float64 * b::AffExpr * c::Float64`. JuMP rewrites as something like
```julia
d::AffExpr = a * b
d = map_coefficients!(d, c)
```
If you swap to `a::Float64 * c::Float64 * b::AffExpr`, then JuMP will do:
```julia
d::Float64 = a * c
e::AffExpr = d * c
```
This saves `N-1` multiplications where `N` is the number of coefficients (in this case, just two(?)), but you get the general idea.

src/model/core/discharge/discharge.jl
```diff
@@ -31,7 +31,7 @@ function discharge!(EP::Model, inputs::Dict, setup::Dict)
     # Variable costs of "generation" for resource "y" during hour "t" = variable O&M
     @expression(EP,
         eCVar_out[y = 1:G, t = 1:T],
-        (inputs["omega"][t]*(var_om_cost_per_mwh(gen[y]) * vP[y, t])))
+        inputs["omega"][t] * var_om_cost_per_mwh(gen[y]) * vP[y, t])
```
odow, line 34: Something here. But the brackets weren't needed? In this case it's better to do `(a * b) * c` than `a * (b * c)`?

src/model/core/co2.jl
lbonaldo: Thank you for clearing this up!

src/model/resources/curtailable_variable_renewable/curtailable_variable_renewable.jl
lbonaldo: Yes, you're right, the two expressions were just transposed versions of each other. Thanks for pointing this out.

RuaridhMacd:
Hi @odow @lbonaldo We're also beginning to see this with Dolphyn. We're going to look into it there too and see if we find the issue.

odow:
Closing in favor of the other PRs

PR #914, opened 2026-06-16
Added Support for Benders Decomposition

dlcole3:
## Description

* This PR adds support for Benders decomposition (single stage only). This was based on work originally done by @filippopecci  who implemented this in other branches, and which this PR uses and extends. 
* Significant refactoring was done to support Benders. a `Benders` flag was added to the settings (and set as `0` as the default). The `generate_model` function was split into a planning level and an operations level. Within the operations level, if the `Benders` flag in settings is set to `1`, it will also call a `capacity_decisions` function which adds copies of the planning level decisions to the subproblems.
* Several modules were refactored to support Benders. For long duration energy storage, hydrogen demand requirements, hourly matching, and CO2 policies (and LDES for VRE + storage modules), there are subperiod and planning level functions that are called when Benders is turned on which place decicision variables on their respective models (e.g., budget variables on the planning level). 
* Gurobi was added as a weak dependency, following a similar pattern as what is used in [MacroEnergy.jl](https://github.com/macroenergy/MacroEnergy.jl). 
* [MacroEnergySolvers.jl](https://github.com/macroenergy/MacroEnergySolvers.jl) was added as a dependency, and its `benders` function is used to solve with Benders decomposition.
* Files were added for writing solutions from Benders decomposition
* Documentation was updated with two new pages of documentation explaining how Benders is set up for GenX, and notes are added to the developer docs on working with Benders.
* A set of tests were added. These tests just solve example cases with and without Benders and test that the solutions are within the Benders gap tolerance (1e-3). 

- [X] Feature
- [ ] Bug Fix
- [X] Documentation Update
- [X] Code Refactor
- [ ] Performance Improvements

## Checklist

- [X] Code changes are sufficiently documented; i.e. new functions contain docstrings and .md files under /docs/src have been updated if necessary.
- [X] The latest changes on the target branch have been incorporated, so that any conflicts are taken care of before merging. This can be accomplished either by merging in the target branch (e.g. 'git merge develop') or by rebasing on top of the target branch (e.g. 'git rebase develop'). Please do not hesitate to reach out to the GenX development team if you need help with this.
- [X] Code has been tested to ensure all functionality works as intended.
- [X] CHANGELOG.md has been updated (if this is a 'notable' change).
- [X] I consent to the release of this PR's code under the GNU General Public license.

## How this can be tested

This PR was primarily tested by running against the example cases in the `example_systems` directory. I ran examples 1-5, 7, 10, and 11 with and without Benders decomposition (with Gurobi as the solver) 

After the PR is approved

- [x] Check that the latest changes on the target branch are incorporated, either via merge or rebase
- [x] Remember to squash and merge if incorporating into develop


dlcole3:
@lbonaldo a couple things to note on the tests in this PR: 
 - The multistage tests failed on some specific Julia versions. I couldn't replicate the problem on my local machine even with the same major Julia version. I believe it is due to numerical differences in the DDP algorithm that is used in that test, such that it was converging to degenerate solutions (plus the subproblem has a pretty high tolerance). Let me know if this is an issue
 - One of the writing outputs tests also failed. Claude suggested this was due to numerical issues in HiGHS and switched to the simplex method over IPM for those tests. 
 - I set up tests for Benders where Benders compares to the monolithic solutions of example cases 1-5, 7, 10, and 11. All of these solve fine on my local machine with Gurobi, but they have numerical issues when using HiGHS. Consequently, I ended up saving the monolithic solutions ahead of time for comparison (some of the monolithic solutions were erroring with no solution), and I had to play with HiGHS parameters to get some of the tests to converge at all. I ended up just using exmaples 1, 4, 5, and 10 for the test script, but it is easy to run the other examples locally using Gurobi.

test/writing_outputs/test_zone_no_resources.jl
```diff
@@ -49,14 +49,21 @@ function test_case()
...
-        "WriteShadowPrices" => 1, 
-        "ParameterScale" => 1)
+        "WriteShadowPrices" => 1)
 
     # Run the case and get the objective value and tolerance
     EP, inputs, _ = redirect_stdout(devnull) do
         run_genx_case_testing(test_path, genx_setup)
     end
-    
+    if result_count(EP) == 0
```
dlcole3, line 58: @lbonaldo The test of Julia 1 with Windows errors on this no resources case used for writing outputs. From what I have looked into, it seems like it's a numerical error in HiGHS, likely due to significant NSE at a high penalty. I opted to use a `@test_broken` line here as I tried many different solver parameters and discussing with Claude and nothing worked in the GitHub workflow test. Open to any thoughts or suggestions you have here

src/benders/benders_planning_problem.jl
```diff
@@ -0,0 +1,176 @@
...
+Raises an error if retrofits or multi-stage planning are enabled, as these are not yet
+compatible with Benders decomposition.
+"""
+function generate_planning_problem(setup::Dict, inputs::Dict, OPTIMIZER::MOI.OptimizerWithAttributes)
+
+    ## Start pre-solve timer
+    presolver_start_time = time()
+    EP = Model(OPTIMIZER)
+
+    #set_string_names_on_creation(EP, Bool(setup["EnableJuMPStringNames"]))
```
lbonaldo: Would it make sense to keep the option to set EnableJuMPStringNames = true for Benders too?

docs/src/User_Guide/benders_decomposition.md
```diff
@@ -0,0 +1,60 @@
...
+
+Note that the stabilization/regularization scheme used in MacroEnergySolvers.jl is turned on when when StabParam is greater than zero. For the regularization scheme to work, the planning problem solver must use an interior point method without crossover. Stabilization is a process in Benders where the master problem can choose less extreme solutions that are on the interior of the feasible set (see [Pecci and Jenkins](https://ieeexplore.ieee.org/abstract/document/10829583)). A challenge of Benders is that, especially at early iterations, the master problem chooses solutions that result in high costs in the subproblems (e.g., at the first iteration, Benders typically chooses to build nothing in the planning level because it is trying to minimize cost without knowledge of the operations) which results in poor cuts. The stabilization scheme generally allows the master to choose solutions that are less extreme and can result in stronger cuts early on.
```
lbonaldo:
```suggestion
Note that the stabilization/regularization scheme used in MacroEnergySolvers.jl is turned on when StabParam is greater than zero. For the regularization scheme to work, the planning problem solver must use an interior point method without crossover. Stabilization is a process in Benders where the master problem can choose less extreme solutions that are on the interior of the feasible set (see [Pecci and Jenkins](https://ieeexplore.ieee.org/abstract/document/10829583)). A challenge of Benders is that, especially at early iterations, the master problem chooses solutions that result in high costs in the subproblems (e.g., at the first iteration, Benders typically chooses to build nothing in the planning level because it is trying to minimize cost without knowledge of the operations) which results in poor cuts. The stabilization scheme generally allows the master to choose solutions that are less extreme and can result in stronger cuts early on.
```

src/benders/benders_planning_problem.jl
```diff
@@ -0,0 +1,176 @@
...
+    EP[:eObj] = AffExpr(0.0)
+
+    planning_model!(EP,setup,inputs)
+
+    @variable(EP,vTHETA[1:inputs["REP_PERIOD"]]>=setup[:ThetaLB])
+
+    ## Define the objective function
+    @objective(EP, Min, setup["ObjScale"]*(EP[:eObj]+sum(vTHETA)))
+
+		@expression(EP, eAvailableCapacity, sum(EP[:eTotalCap]))
```
lbonaldo:
```suggestion
	@expression(EP, eAvailableCapacity, sum(EP[:eTotalCap]))
```

src/benders/benders_utility.jl
```diff
@@ -0,0 +1,125 @@
...
+subperiod `w`.  Each sub-dictionary also carries `REP_PERIOD = 1` and a `SubPeriod` field
+with its index, making it self-contained for building a single operational subproblem.
+"""
+function separate_inputs_subperiods(inputs::Dict)
+
+    inputs_all=Dict();
+    number_periods = inputs["REP_PERIOD"];
+    hours_per_subperiod = inputs["hours_per_subperiod"];
+    
+    ####### entries_to_be_changed = ["omega","REP_PERIOD",","INTERIOR_SUBPERIODS","START_SUBPERIODS","pP_Max","T","fuel_costs","Weights","pD","C_Start"];
```
lbonaldo:
```suggestion
```

src/benders/benders_utility.jl
```diff
@@ -0,0 +1,125 @@
...
+
+    for w in 1:number_periods
+        inputs_all[w] = deepcopy(inputs);
```
lbonaldo: Is this deepcopy necessary? If I'm understanding the logic correctly, could we loop over the keys in the dictionary and only copy slices of the time series?

src/benders/benders_utility.jl
```diff
@@ -0,0 +1,125 @@
...
+- `"planning_problem"`: the master JuMP model
+- `"planning_variables"`: names of first-stage decision variables
+- `"subproblems"`: operational subproblem dicts — a `Vector{Dict}` when running with a
+  single Julia process (`nworkers() == 1`), or a `DArray` across multiple workers otherwise
+- `"planning_variables_sub"`: per-subperiod mapping of linking variable names
+
+Using a `Vector{Dict}` when only one process is available avoids routing every subproblem
+solve through Julia's distributed message-passing infrastructure (`@fetchfrom 1 / @spawnat 1`),
+which can deadlock when the solver (e.g. HiGHS IPM) spawns OpenMP threads that interfere
+with Julia's cooperative task scheduler on the single OS thread.
```
lbonaldo: I'm not sure I fully understand the comment here

src/benders/benders_utility.jl
```diff
@@ -0,0 +1,125 @@
...
+
+Checks `eTotalCap`, and optionally `eTotalCapEnergy`, `eTotalCapCharge`, and
+`eAvail_Trans_Cap` when those keys are present in the model.
+"""
+function check_negative_capacities(EP::Model)
+
+	neg_cap_bool = false;
+	tol = -1e-8;
+	if any(value.(EP[:eTotalCap]).< tol) 
+			neg_cap_bool = true;
```
lbonaldo: Probably not necessary, as this code doesn't seem to be referenced anywhere, but two comments: 
1. I think we should convert the following elseifs into independent if-else statements. For example, if the model has eTotalCapEnergy, but it's positive (or zero), the code will skip the checks on eTotalCapCharge and eAvail_Trans_Cap.  
2.  to improve performance, we could `return` after any of the conditions is met (as we don't probably need to check other conditions if one evaluates to true).

src/case_runners/case_runner.jl
```diff
@@ -207,3 +234,57 @@ function run_genx_case_multistage!(case::AbstractString, mysetup::Dict, optimize
...
+            cluster_inputs(case, settings_path, mysetup)
+        else
+            println("Time Series Data Already Clustered.")
+        end
+    end
+    mysetup["settings_path"] = settings_path;
+
+    myinputs = load_inputs(mysetup, case);
+
+    # SPLIT BENDERS IF NOT USING TDR
```
lbonaldo: Is this comment still applicable?

src/model/resources/storage/long_duration_storage.jl
```diff
@@ -259,3 +259,196 @@ function long_duration_storage!(EP::Model, inputs::Dict, setup::Dict)
...
+    #             (1-self_discharge(gen[y]))*vSOCw[y,r]-(1/efficiency_down(gen[y])*vP[y,hours_per_subperiod*(dfPeriodMap[r,:Rep_Period_Index]-1)+1])
+    #             +(efficiency_up(gen[y])*vCHARGE[y,hours_per_subperiod*(dfPeriodMap[r,:Rep_Period_Index]-1)+1])
+    #             +vdSOC_maxPos[y,dfPeriodMap[r,:Rep_Period_Index]] <= eTotalCapEnergy[y])
+
+    #     # Min storage content within each modeled period cannot be negative
+    #     @constraint(EP, cSoCLongDurationStorageMinInt[y in STOR_LONG_DURATION, r in NON_REP_PERIODS_INDEX],
+    #             (1-self_discharge(gen[y]))*vSOCw[y,r]-(1/efficiency_down(gen[y])*vP[y,hours_per_subperiod*(dfPeriodMap[r,:Rep_Period_Index]-1)+1])
+    #             +(efficiency_up(gen[y])*vCHARGE[y,hours_per_subperiod*(dfPeriodMap[r,:Rep_Period_Index]-1)+1])
+    #             +vdSOC_maxNeg[y,dfPeriodMap[r,:Rep_Period_Index]] >= 0)
+    # end
```
lbonaldo: It would be probably be nice to warn the users that these constraints (setup["LDSAdditionalConstraints"] == 1) are not currently available with Benders

src/model/generate_model.jl
```diff
@@ -74,6 +74,331 @@ function generate_model(setup::Dict, inputs::Dict, OPTIMIZER::MOI.OptimizerWithA
...
     presolver_start_time = time()
 
+    # Generate Energy Portfolio (EP) Model
+    EP = if Bool(setup["EnableJuMPDirectModel"])
+        opt_instance = MOI.instantiate(OPTIMIZER)
+        direct_model(opt_instance)
+    else
+        Model(OPTIMIZER)
+    end
+    #set_string_names_on_creation(EP, Bool(setup["EnableJuMPStringNames"]))
```
lbonaldo: Same as above

src/write_outputs/write_benders_output.jl
```diff
@@ -0,0 +1,1761 @@
...
+			println("Time elapsed for writing shutdown is")
+			println(elapsed)
+		end
+	end
+
+	# Planning-problem dual outputs (subsidy revenue, CO2 cap prices) are intentionally not
+	# written for Benders: these shadow prices are not available from the incumbent planning
+	# solution and are not reliably recoverable from the cut-laden master problem. They can be
+	# reintroduced once a trustworthy planning dual is available.
+end
```
lbonaldo: I think it would be nice to also notify the users about this, as they might not see this comment on the source code

src/write_outputs/write_planning_problem_costs.jl
```diff
@@ -0,0 +1,91 @@
...
+    var_vals = planning_sol.values  # Dict{String, Float64}
+    val_fn = var -> get(var_vals, name(var), 0.0)
+    EP = planning_problem
+
+    cost_list = [
+        "cTotal",
+        "cFix",
+        "cNetworkExp",
+        "cUnmetPlanningPolicyPenalty",
+    ]
```
lbonaldo: I see that variable cost, startup and fuel cost and nse cost are not reported here. How hard do you think it would be to write them too?

src/model/generate_model.jl
```diff
@@ -74,6 +74,331 @@ function generate_model(setup::Dict, inputs::Dict, OPTIMIZER::MOI.OptimizerWithA
...
+    # demand = generation + storage discharge - storage charge - demand deferral + deferred demand satisfaction - demand curtailment (NSE)
+    #          + incoming power flows - outgoing power flows - flow losses - charge of heat storage + generation from NACC
+    @constraint(EP,
+        cPowerBalance[t = 1:T, z = 1:Z],
+        EP[:ePowerBalance][t, z]==inputs["pD"][t, z])
+end
+
+
+
+function generate_model_legacy(setup::Dict, inputs::Dict, OPTIMIZER::MOI.OptimizerWithAttributes)
```
lbonaldo: i think we can remove this function as nothing reference it anymore.

src/model/generate_model.jl
```diff
@@ -74,6 +74,331 @@ function generate_model(setup::Dict, inputs::Dict, OPTIMIZER::MOI.OptimizerWithA
...
+            energy_share_requirement_subperiod!(EP, inputs, setup)
+        else
+            energy_share_requirement!(EP, inputs, setup)
+        end
+    end
+
+    # Hourly Matching Requirement
+    if setup["HourlyMatchingRequirement"] == 1
+        hourly_matching!(EP, inputs)
+    end
```
lbonaldo, line 376: do we need to split this block into a Benders and a non-Benders version?

src/model/policies/hourly_matching.jl
```diff
@@ -58,3 +58,69 @@ function hourly_matching!(EP::Model, inputs::Dict)
 
     @constraint(EP, cHourlyMatching[t = 1:T, hm = 1:nHM], EP[:eHM][t, hm] >= EP[:eHMDemand][t, hm])
 end
+
+function hourly_matching_planning!(EP::Model, inputs::Dict)
+    nHM = inputs["nHM"]
+    t = inputs["T"] # includes all time points of system
```
lbonaldo: I feel like this is a typo
```suggestion
    T = inputs["T"] # includes all time points of system
```

src/model/policies/hourly_matching.jl
```diff
@@ -58,3 +58,69 @@ function hourly_matching!(EP::Model, inputs::Dict)
...
+    println("Hourly Matching Policies Module")
+    T = inputs["T"]
+    nHM = inputs["nHM"]
+    w = inputs["SubPeriod"];
+
+    @variable(EP, vHMShortFallBudget[[w], hm = 1:nHM]>=0)
+    
+    # Construct hourly matching demand by adding absolute profiles with zonal percentages, if applicable
+    @expression(EP, eHMDemand[t = 1:T, hm = 1:nHM],
+        inputs["dfHM_absolute"][t, hm])
```
lbonaldo, line 94: do we need to slice dfHM_absolute too?

src/write_outputs/write_benders_output.jl
```diff
@@ -0,0 +1,1761 @@
...
+		if get(output_settings_d, "WriteStatus", true)
+			write_status(outpath, inputs, setup, planning_problem)
+		end
+		@warn "Benders planning problem has no solver values in the model object; skipping detailed (operational) output files. Use benders_results fields for algorithm diagnostics."
+		return nothing
+	end
+
+	#TODO: Check that Benders converged;
+
+    CSV.write(joinpath(outpath, "benders_convergence.csv"),dfConv)
```
lbonaldo: shall we write this file before the check !has_values? The docstring says that "If the planning problem has no solver values (e.g. the algorithm did not converge), only the status file and convergence CSV are written and the function returns early." however, this write call happens after the return

src/benders/benders_planning_problem.jl
```diff
@@ -0,0 +1,176 @@
...
+    #set_string_names_on_creation(EP, Bool(setup["EnableJuMPStringNames"]))
+    # Introduce dummy variable fixed to zero to ensure that expressions like eTotalCap,
+    # eTotalCapCharge, eTotalCapEnergy and eAvail_Trans_Cap all have a JuMP variable
+    @variable(EP, vZERO==0)
+
+	if !isempty(inputs["RETROFIT_OPTIONS"])
+		error("Benders not yet supported with retrofits")
+	elseif setup["MultiStage"] > 0
+		error("Multistage and Benders are not integrated yet.")
+	end
```
lbonaldo, line 37: I can see that the MGA function is not being called inside the Benders case runner. Should we maybe notify the users here as we do for retrofit and multistage? 
How about UCommit == 1?

src/model/capacity_decisions.jl
```diff
@@ -0,0 +1,694 @@
...
+        ### VARIABLES ###
+        @variables(EP, begin
+            vCAPCHARGE_AC[y in NEW_CAP_CHARGE_AC] >= 0               # Charge capacity AC component built for VRE storage [MW]
+            vRETCAPCHARGE_AC[y in RET_CAP_CHARGE_AC] >= 0            # Charge capacity AC component retired for VRE storage [MW]
+        end)
+
+        # DEV NOTES: Uncomment when adding Multistage support
+        if MultiStage == 1
+            @variable(EP, vEXISTINGCAPCHARGEAC[y in VS_ASYM_AC_CHARGE]>=0)
+        end
```
lbonaldo: should we comment this block out as well?

src/model/core/ldes_slack.jl
```diff
@@ -0,0 +1,55 @@
...
+    
+    PenaltyValue = 100*(inputs["Weights"]/inputs["H"])*inputs["Voll"][1] ;
+    println("LDES slack penalty value is:")
+    println(PenaltyValue)
+
+	@expression(EP,eObjSlack,sum(PenaltyValue[w]*vLDS_SLACK_MAX[w] for w in 1:inputs["REP_PERIOD"]))
+    
+    EP[:eObj] += eObjSlack
+
+    if setup["LDES_Feasible"]==1
```
lbonaldo: Should we document this flag too in the documentation?

src/write_outputs/write_status.jl
```diff
@@ -7,15 +7,19 @@ function write_status(path::AbstractString, inputs::Dict, setup::Dict, EP::Model
...
     # https://jump.dev/MathOptInterface.jl/v0.9.10/apireference/#MathOptInterface.TerminationStatusCode
     status = termination_status(EP)
+    has_solution = has_values(EP)
+    objval = has_solution ? objective_value(EP) : missing
 
     # Note: Gurobi excludes constants from solver reported objective function value - MIPGap calculated may be erroneous
     if (setup["UCommit"] == 0 || setup["UCommit"] == 2)
-        dfStatus = DataFrame(Status = status, Solve = inputs["solve_time"],
-            Objval = objective_value(EP))
+        dfStatus = DataFrame(Status = status,
```
lbonaldo: Can we maybe keep the solve time output if present in the dictionary (e.g. for monolithic runs)?

src/GenX.jl
```diff
@@ -38,6 +38,10 @@ using RecursiveArrayTools
 using Statistics
 using HiGHS
 using Logging
+using MacroEnergySolvers
+using Distributed
+using DistributedArrays
+using ClusterManagers
```
lbonaldo: Is this dependency still used?

Project.toml
```diff
@@ -5,16 +5,22 @@ version = "0.4.6"
...
+Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
+DistributedArrays = "aaf54ef3-cdf8-58ed-94cc-d582ad619b94"
 HiGHS = "87dc4568-4c63-4d18-b0c0-bb2238e4078b"
+JLD2 = "033835bb-8acc-5ee8-8aae-3f567f8a3819"
 JuMP = "4076af6c-e467-56ae-b986-b466b2749572"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
+LoggingExtras = "e6f89c97-d47a-5376-807f-9c37f3926c36"
+MacroEnergySolvers = "466a4f75-6c9c-4ef2-97df-f5f43ac00c23"
 MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
```
lbonaldo: I think we can drop JLD2, LoggingExtras and ClusterManagers here. We might also need to add [compat] entries for Distributed, DistributedArrays and Gurobi so that the future release will pass the AutoMerge checks.

src/model/generate_model.jl
```diff
@@ -74,6 +74,331 @@ function generate_model(setup::Dict, inputs::Dict, OPTIMIZER::MOI.OptimizerWithA
...
+
+    # Initialize Objective Function Expression
+    EP[:eObj] = AffExpr(0.0)
+
+    # Guard against unsupported Benders + VRE-STOR combination
+    if setup["Benders"] == 1 
+        if setup["LDSAdditionalConstraints"] == 1
+            @warn "LDSAdditionalConstraints=1 applies to problems with non-representative periods and is not supported in the current Benders implementation. Benders will proceed as if LDSAdditionalConstraints=0"
+        end
+    end
```
lbonaldo: Is this comment/block still applicable?

src/model/generate_model.jl
```diff
@@ -74,6 +74,331 @@ function generate_model(setup::Dict, inputs::Dict, OPTIMIZER::MOI.OptimizerWithA
...
+        flexible_demand!(EP, inputs, setup)
+    end
+
+    # Model constraints, variables, expression related to thermal resource technologies
+    if !isempty(inputs["THERM_ALL"])
+        thermal!(EP, inputs, setup)
+    end
+
+    # Model constraints, variables, expressions related to the co-located VRE-storage resources
+    # (Benders case with VRE_STOR already errored at generate_model entry point)
```
lbonaldo: Same as above

src/model/capacity_decisions.jl
```diff
@@ -0,0 +1,694 @@
...
+    
+    VRE_STOR = inputs["VRE_STOR"] # Set of VRE-STOR resources
+    # MultiStage = setup["MultiStage"]
+
+    # Make sure to set total cap vars
+    gen = inputs["RESOURCES"]
+
+    G = inputs["G"] # Number of resources (generators, storage, DR, and DERs)
+    T = inputs["T"] # Number of time steps (hours)
+    Z = inputs["Z"] # Number of zones
```
lbonaldo: These seems to be unused?

src/model/capacity_decisions.jl
```diff
@@ -0,0 +1,694 @@
...
+    VS_SYM_DC = inputs["VS_SYM_DC"]
+    VS_SYM_AC = inputs["VS_SYM_AC"]
+    VS_LDS = inputs["VS_LDS"]
+
+    START_SUBPERIODS = inputs["START_SUBPERIODS"]
+    INTERIOR_SUBPERIODS = inputs["INTERIOR_SUBPERIODS"]
+    hours_per_subperiod = inputs["hours_per_subperiod"]     # total number of hours per subperiod
+    rep_periods = inputs["REP_PERIOD"]
+
+    MultiStage = setup["MultiStage"]
```
lbonaldo: Maybe also these flags are unused?

docs/make.jl
```diff
@@ -35,18 +35,20 @@ pages = OrderedDict(
         "Model Inputs" => "User_Guide/model_input.md",
         "Time-domain Reduction Inputs" => "User_Guide/TDR_input.md",
         "Running the Time-domain Reduction" => "User_Guide/running_TDR.md",
+        "Benders Decompositiion" => "User_Guide/benders_decomposition.md",
```
lbonaldo:
```suggestion
        "Benders Decomposition" => "User_Guide/benders_decomposition.md",
```

test/test_benders_vs_monolithic.jl
```diff
@@ -0,0 +1,131 @@
...
+# Pre-clustered TDR data lives in test/benders/<case_name>/TDR_results/.
+const EXAMPLE_CASES = [
+    (1,  "1_three_zones"),
+    # (2,  "2_three_zones_w_electrolyzer_and_hourly_matching"),
+    # (3,  "3_three_zones_w_co2_capture"),
+    (4,  "4_three_zones_w_policies_slack"),
+    (5,  "5_three_zones_w_piecewise_fuel"),
+    # (7,  "7_three_zones_w_colocated_VRE_storage"),
+    (10, "10_IEEE_9_bus_DC_OPF"),
+]
```
lbonaldo, line 42: Could we maybe remove the other examples if they are not being used anymore?

src/benders/benders_subproblems.jl
```diff
@@ -0,0 +1,249 @@
...
+Delegates to `MacroEnergySolvers.solve_subproblems` using the planning variable values
+in `results.planning_sol`, appending the subproblem solution as `subop_sol` in the
+returned `results` NamedTuple.  Operates on either a local `Vector{Dict}` or a
+distributed `DArray` of subproblem dicts.
+"""
+function update_with_subproblem_solutions!(subproblems::Union{Vector{Dict{Any, Any}},DistributedArrays.DArray}, results::NamedTuple)
+
+    subop_sol = MacroEnergySolvers.solve_subproblems(subproblems, results.planning_sol, true)
+
+    results = (; results..., subop_sol = subop_sol)
```
lbonaldo: Since results is a namedtuple (immutable), I don't think it gets updated in place. As a result, this function doesn't mutate results, but only subproblems (correct).

src/benders/benders_planning_problem.jl
```diff
@@ -0,0 +1,176 @@
...
+
+# Private helper: dispatch to the appropriate solver-specific configure function.
+const _BENDERS_CONFIGURE_FUNCTIONS = Dict{String, Function}(
+    "highs"  => configure_highs,
+    "gurobi" => configure_gurobi,
+    "cplex"  => configure_cplex,
+    "clp"    => configure_clp,
+    "cbc"    => configure_cbc,
+    "scip"   => configure_scip,
+)
```
lbonaldo: It would probably be nice to use this variable in configure_solver too to improve DRY

src/write_outputs/write_benders_output.jl
```diff
@@ -0,0 +1,1761 @@
...
+	end
+
+	if get(output_settings_d, "WriteCosts", true)
+		write_planning_problem_costs(outpath, inputs, setup, benders_results, planning_problem)
+	end
+
+	if !has_values(planning_problem)
+		if get(output_settings_d, "WriteStatus", true)
+			write_status(outpath, inputs, setup, planning_problem)
+		end
```
lbonaldo: are these lines the same as 25-27?

lbonaldo (commented):
Hey @dlcole3! Thank you so much for this PR! This is an incredible addition to GenX! I completed a first round of review, and I added a few comments on the code as i was reading it. Most of them are really minor comments, the only one that might be more important is the one related to the hourly matching constraint. Of course let me know if you have any questions.

src/benders/benders_planning_problem.jl
dlcole3: I am setting this for both the planning and subproblems.

src/benders/benders_utility.jl
dlcole3: I agree with you. I think there is a more efficient way of doing this. I will have it do all the time slices first and then loop through the keys of the inputs dictionary and copy over everything that wasn't in the time slices.

src/model/generate_model.jl
dlcole3: I updated this one to allow string names here.

src/benders/benders_utility.jl
dlcole3: I updated the phrasing a little bit: 

```
The "subproblems" entry of the dictionary uses a `Vector{Dict}` when only one process is 
available to avoid routing every subproblem solve through Julia's distributed message-passing
infrastructure (`@fetchfrom 1 / @spawnat 1`), which can deadlock when the solver (e.g. HiGHS IPM) 
spawns OpenMP threads that interfere with Julia's cooperative task scheduler on the single OS thread. 
```

There were some issues in the unit tests with Windows where it would just stall out. Claude suggested it was because of conflicts with threading in HiGHS.

src/benders/benders_utility.jl
dlcole3: I am deleting these lines. You are right that it never gets called anywhere. This is a residual function that I missed that was used before MacroEnergySolvers.

src/case_runners/case_runner.jl
dlcole3: Nope, deleting this comment.

src/model/resources/storage/long_duration_storage.jl
dlcole3: There is a warning for this inside generate_model.

src/write_outputs/write_benders_output.jl
dlcole3: I converted this into a warn statement.

src/write_outputs/write_planning_problem_costs.jl
dlcole3: Fuel and NSE costs are written already by the "write_benders_output.jl" file because they are in the subproblems of Benders. I will check on variable O&M as well to see if/how it gets written and make sure Benders is outputting it too

src/model/policies/hourly_matching.jl
dlcole3: Yep, good catch. I will update that.

src/write_outputs/write_benders_output.jl
dlcole3: Good catch. I will update this.

src/model/capacity_decisions.jl
dlcole3: Yep. Done.

src/GenX.jl
dlcole3: I don't believe so. I will remove it.

src/write_outputs/write_planning_problem_costs.jl
dlcole3: Okay, I see what you mean with having them in the planning section. I will update to include these costs.

src/benders/benders_planning_problem.jl
dlcole3: Good catch. Done.

src/model/generate_model.jl
dlcole3: No, the coment should be updated.

dlcole3: Actually, this whole block should be moved into the operation model. Right now, it is in `generate_model` which won't run in the case that "Benders" == 1. I will move this to the operations model function instead.

src/model/capacity_decisions.jl
dlcole3: They are. I just deleted them (plus an extra `T = inputs["T"]` in the following function)

src/model/capacity_decisions.jl
dlcole3: You are correct. I just deleted them.

test/test_benders_vs_monolithic.jl
dlcole3: Yep. Just deleted them. Also deleted the earlier `KNOWN_OPTIMA` dict entries in the same file.

src/benders/benders_planning_problem.jl
dlcole3: Actually, the Benders implementation requires the string names and uses that to map linking variables between the planning and subproblems. I will add a line here to warn the user if they set this to false that it won't do anything

src/model/generate_model.jl
dlcole3: Yeah, that second line is old and I just deleted it.

docs/make.jl
dlcole3: Thanks. I just corrected this on my local that I will push shortly.

src/model/core/ldes_slack.jl
dlcole3: I updated the documentation. I actually put this LDES_Feasible flag outside of the function. I don't see a reason to add the slacks if we just fix them to zero, so these slacks are only added if LDES_Feasible == 0.

src/benders/benders_subproblems.jl
dlcole3: I removed this function so the case runner now just calls the following instead of having a special function for this. This should update `results` in place, which I think is what we want so taht subop_sol gets reset in the named tuple. 

```julia

    subop_sol = MacroEnergySolvers.solve_subproblems(subproblems, results.planning_sol, true)
    
    results = (; results..., subop_sol = subop_sol)
```

src/benders/benders_planning_problem.jl
dlcole3: I went ahead and replaced the functions [here](https://github.com/GenXProject/GenX.jl/blob/3d9596ad9ce69e66ed0404754e58250059bf612b/src/configure_solver/configure_solver.jl#L28-L34) with a constant called _SOLVER_CONFIGURE_FUNCTIONS which is available for both the `configure_solver` function and the Benders functions.

src/write_outputs/write_benders_output.jl
dlcole3: Yes. these are identical. I will delete the second one on lines 64-66.

src/write_outputs/write_status.jl
dlcole3: Yeah, I added this back in. I think it makes sense to just have a separate write_status function for Benders so that it can check the convergence tolerance and other parameters. The above version tests `termination_status` of the planning model, but I think this should output the overall results with the total solve time. It also now notes the termination status of the Benders algorithm rather than the planning problem.

src/model/generate_model.jl
dlcole3: There was already code for this, but I forgot to call it in this code. Running the example cases actually never caught it because it appears virtually all time points have the same values, so for testing on my local machine, I altered example 2 for testing by changing the hourly matching requirement at different time points and it worked okay.

Project.toml
dlcole3: I updated the code for setting the processors to check the slurm environment like Macro does, so it now needs SlurmClusterManager. But I dropped JLD2 and LoggingExtras and added to the compat section.

dlcole3:
@lbonaldo In adding a compat note for Gurobi as a weak dependency, the Julia 1.6 test broke. I updated the workflow to drop the 1.6 test and I updated the Project.toml in the last commit to require Julia 1.9 or above. This is consistent with macro as well. Please let me know if this is an issue.

dlcole3:
Some of the previous tests with Benders using Julia 1.10 and above were hanging, but version 1.9 would work fine. After some disucssion with Claude, it seems that it is related to an issue in HiGHS on newer versions. Claude made a Project.toml for the tests so that they are restricted to an earlier version of HiGHS, and the tests now pass more consistently.

lbonaldo (approved):
This PR looks good to me. Thank you so much for addressing all the comments, you did an incredible work @dlcole3!

Issue #900, opened 2026-01-26
[Question]: DC-OPF–Based Transmission Expansion in GenX?


M0-Adnan:
### Question

Hi everyone,

In previous GenX versions, the transmission expansion option is only available with the transport model. I would like to ask whether there is currently an option to consider transmission expansion with DC OPF, or if the group has any plans to implement a DC-OPF–based transmission expansion.

Thanks!

dlcole3:
@M0-Adnan We are working on supporting this. See #915.

PR #915, opened 2026-07-15
Added Support for DCOPF Constrained Expansion

dlcole3:
## Description

This PR adds code to support expansion decisions of DCOPF constrained GenX models that can be solved both monolithically or using Benders Decomposition. This work was originally started by @sambuddhac. It includes updates to settings, inputs loading, model construction, and solution methods. Part of the motivation for this work is supporting nodal spatial resolution of CEMs. In these cases, you might solve with DCOPF constraints, and you might also want to have integer decisions on new transmisison or generation builds (e.g., CC might be built in 500 MW increments to avoid unrealistic modular CC construction in lots of different zones). This PR is intended as a second step after the Benders PR (#914) is merged.
 * There is now a `DiscreteInvestments` flag in the settings of GenX. This controls whether there can be integer variables in the model (not that I chose to call this `DiscreteInvestments` over `IntegerInvestments` because MacroEnergySolvers.jl already has an `IntegerInvestments` setting). This can be applied to both resources and transmission lines. If `DiscreteInvestments` is set to 1, then GenX will look for a column header `Discrete_Build` in the Network.csv, Thermal.csv, and Vre.CSV and throw a warning statement if none of these files have the column header. This header controls which resources or lines can have discrete builds. This also gives the user flexiblity in controlling which components should be treated as integer/binary build decisions.
 *  DC-OPF constraints now support expansion and discrete new builds. Previously, DC-OPF was only applied to fixed, existing networks. Now, there are model constraints permitting line expansion as well as new lines on existing or new corridors. 
 * Network.csv supports additional column names. If the `Discrete_Build` column has a 1, then there can be at least one new line constructed on that corridor. GenX internally builds a new DataFrame with each line separated out with individual flow variables on them. This is important for handling DC-OPF constraints. In addition, it supports the case where there may be multiple discrete parallel lines on a corridor, and the user can pass these directly with only one line of the CSV. There is an additional constraint added which adds a solve order to identical lines to avoid degeneracy.
 *  Line expansion supports both expansion and discrete new builds in the same model (though new lines cannot be expanded). If a user sets `Discrete_Build` to zero, but there is a nonzero `Line_Max_Reinforcement_MW`, then that line will be treated as the traditional line expansion available in GenX (note that this can essentially be thought of as reconductoring). If the discrete build flag is set to 1, then `Line_Max_Reinforcement_MW` is used to determine the size of the new lines (note: we can restructure this if you think new lines should have more columns to support this instead of using the existing), with the `New_Line_Cap_Size_MW` column name used to determine how may lines fit into `Line_Max_Reinforcement_MW`. 
 * When DC_OPF is on with discrete builds, lines are placed into different sets inside the inputs dictionary internally. These are `FIXED_LINES` which are lines that can't be expanded, `EXPANSION_LINES`, which are lines that have continuous reinforcement on them, `DISCRETE_BUILD_LINES` which are lines that have a binary build decision associated with them, and `PHANTOM_LINES`, which are the `DISCRETE_BUILD_LINES` which are on new corridors (this is important for the line angle limit constraints, which you only want to have applied if the line is built). 
 * There is a new example case 12 that is a copy of example case 10 but with line data adjusted to support new builds. 
 * There are new tests that example case 12 (but with a smaller time horizon) for testing purposes, both with and without Benders
 * Benders decomposition has been extended to work with discrete build decisions following the methods outlined in [this paper](https://arxiv.org/abs/2603.29867). This includes a transport model and LP hotstarting. MacroEnergySolvers.jl supports the LP relaxation of binary decisions, but I had two concerns with using its methods and defined my own in GenX: 1) I want the flexibility to turn off the stabilization method after hot starting is finished, and 2) binary varaibles do not get an explicit upper bound set on them inside MacroEnergySolvers which I think can cause issues for the big M constraints here. 
 * The documentation has been updated to describe how these new features can be used

A couple of additional notes: 
 * Line losses have not been tested in this PR. I believe that they should still work because all the lines use the `vFLOW` variable, and this will be zero for any line not constructed. @sambuddhac may have a better idea on whether this is the case.
 * Multistage runs are not explicitly handled in this PR. 
 * The Changelog has been updated, but it is a continuation of the former Benders PR, so the updates are at the end of the Benders work. Not sure if the Benders PR gets merged if we will need to update that Changelog
 * For problems with lots of nodes/zones, GenX can be quite slow. There is a need for future model construction improvements to handle problems with large zones. This is planned and will follow up this PR.

@gschivley and I had discussions on the above structuring, so tagging him in case he wants to give input on how it came together. 

- [X] Feature
- [ ] Bug Fix
- [X] Documentation Update
- [ ] Code Refactor
- [ ] Performance Improvements

## Related Tickets & Documents

This PR also address the feature request of #900 

## Checklist

- [X] Code changes are sufficiently documented; i.e. new functions contain docstrings and .md files under /docs/src have been updated if necessary.
- [ ] The latest changes on the target branch have been incorporated, so that any conflicts are taken care of before merging. This can be accomplished either by merging in the target branch (e.g. 'git merge develop') or by rebasing on top of the target branch (e.g. 'git rebase develop'). Please do not hesitate to reach out to the GenX development team if you need help with this.
- [X] Code has been tested to ensure all functionality works as intended.
- [X] CHANGELOG.md has been updated (if this is a 'notable' change).
- [X] I consent to the release of this PR's code under the GNU General Public license.

## How this can be tested

There is a new example case (12) and a new set of unit tests. The unit tests run a smaller time horizon version of example 12 with both monolithic and Benders. In addition, the unit test running the three zone case also runs it with `DiscreteInvestments=1` to test integer build decisions on resources. 

After the PR is approved

- [ ] Check that the latest changes on the target branch are incorporated, either via merge or rebase
- [ ] Remember to squash and merge if incorporating into develop


example_systems/2_three_zones_w_electrolyzer_and_hourly_matching/Run.jl
```diff
@@ -1,3 +1,4 @@
 using GenX
+using Gurobi
 
-run_genx_case!(dirname(@__FILE__))
+run_genx_case!(dirname(@__FILE__), Gurobi.Optimizer)
```
lbonaldo, line 4: I guess we can cleanup this file here, am I right?

src/case_runners/case_runner.jl
```diff
@@ -42,6 +42,40 @@ function run_genx_case!(case::AbstractString, optimizer::Any = HiGHS.Optimizer)
             mysetup_benders = configure_benders(benders_settings_path)
             mysetup = merge(mysetup, mysetup_benders)
 
+            if mysetup["DC_OPF"] != 1 && mysetup[:RunTransportModel]
+                @warn "`RunTransportModel` only works with DC_OPF = 1 but DC_OPF is set to $(mysetup["DC_OPF"]). Turning off `RunTransportModel`"
```
lbonaldo, line 46: do we need to turn off `RunTransportModel`?

src/case_runners/case_runner.jl
```diff
@@ -42,6 +42,40 @@ function run_genx_case!(case::AbstractString, optimizer::Any = HiGHS.Optimizer)
...
+                        Distributed.remotecall_eval(Main, workers(), :(using GenX))
+                    elseif current > target
+                        @warn "Benders: $current workers are already running but NWorkers=$target was requested. Proceeding with $current workers."
+                    else
+                        @info "Benders: $current workers already running — no additional workers needed."
+                    end
+                else
+                    @warn "Benders: Distributed=true but NWorkers=$target (must be > 1 to enable parallel solving). Running sequentially."
+                end
+            end
```
lbonaldo, line 77: isn't this part also in `setup_benders_workers!`?

src/configure_solver/configure_benders.jl
```diff
@@ -26,16 +30,20 @@ function configure_benders(settings_path::String)
...
-        :ThetaLB                   => 0.0,
+        :ConvTol                      => 1e-3,
+        :MaxIter                      => 50,
+        :MaxCpuTime                   => 7200,
+        :StabParam                    => 0.0,
+        :StabDynamic                  => false,
+        :ExpectFeasibleSubproblems    => false,
+        :IntegerInvestment            => false,
+        :Distributed                  => false,
+        :NWorkers                     => 0,
```
lbonaldo, line 41: This default value changes from -1 to 0. Is that okay?

src/model/resources/vre_stor/vre_stor.jl
```diff
@@ -621,6 +587,17 @@ function stor_vre_stor!(EP::Model, inputs::Dict, setup::Dict)
...
+        VS_ASYM_DC_CHARGE = inputs["VS_ASYM_DC_CHARGE"]
+        VS_ASYM_AC_CHARGE = inputs["VS_ASYM_AC_CHARGE"]
+        VS_ASYM_DC_DISCHARGE = inputs["VS_ASYM_DC_DISCHARGE"]
+        VS_ASYM_AC_DISCHARGE = inputs["VS_ASYM_AC_DISCHARGE"]
+
+        # Constraint 4: Charging + Discharging AC Maximum: see main module because capacity reserve margin/operating reserves may alter constraint
+        @expression(EP, eChargeDischargeMaxAC[y in VS_SYM_AC, t = 1:T],
+            EP[:vP_AC_DISCHARGE][y, t]+EP[:vP_AC_CHARGE][y, t])
+    end
+
```
lbonaldo, line 600: Is this part related to this PR?

src/model/core/discharge/investment_discharge.jl
```diff
@@ -116,7 +129,7 @@ function investment_discharge!(EP::Model, inputs::Dict, setup::Dict)
     ### Need editting ##
     @expression(EP, eCFix[y in 1:G],
         if y in NEW_CAP # Resources eligible for new capacity (Non-Retrofit)
-            if y in COMMIT
+            if y in CAP_SIZE_SCALED
```
lbonaldo, line 132: isn't it this part also in capacity_decision.jl?

src/model/resources/vre_stor/vre_stor.jl
```diff
@@ -1,52 +1,18 @@
...
-    and charge of battery) are equal to the technology's total discharging to and charging from the grid:
+    vre_stor!(EP::Model, inputs::Dict, setup::Dict)
+
+Operational-stage coordinator for the VRE-STOR module.
+
+Builds and links component operational blocks (`inverter_vre_stor!`, `solar_vre_stor!`,
+`wind_vre_stor!`, `stor_vre_stor!`, `elec_vre_stor!`) and enforces module-level
+constraints using planning capacities from investment functions.
+
+Core constraints include AC-balance dispatch and grid-interface limits:
```
lbonaldo, line 10: are we removing the docstring for vre_stor here?

src/model/core/transmission/investment_transmission.jl
```diff
@@ -46,6 +50,9 @@ function investment_transmission!(EP::Model, inputs::Dict, setup::Dict)
     if NetworkExpansion == 1
         # Transmission network capacity reinforcements per line
         @variable(EP, vNEW_TRANS_CAP[l in EXPANSION_LINES]>=0)
+        if DiscreteInvestments == 1
+            @variable(EP, vNEW_TRANS_LINES[l in DISCRETE_BUILD_LINES], Bin)
+        end
```
lbonaldo, line 55: Same as above. I think this part is also in capacity_decision.jl

src/model/generate_model.jl
```diff
@@ -81,11 +81,17 @@ function generate_model(setup::Dict, inputs::Dict, OPTIMIZER::MOI.OptimizerWithA
...
 
     # Initialize Objective Function Expression
     EP[:eObj] = AffExpr(0.0)
 
+    # Guard against unsupported Benders + VRE-STOR combination
+    if setup["Benders"] == 1 
+        if setup["LDSAdditionalConstraints"] == 1
+            @warn "LDSAdditionalConstraints=1 applies to problems with non-representative periods and is not supported in the current Benders implementation. Benders will proceed as if LDSAdditionalConstraints=0"
+        end
+    end
```
lbonaldo, line 94: are these lines from the Benders PR?

src/model/generate_model.jl
```diff
@@ -368,11 +371,7 @@ function operation_model!(EP::Model, setup::Dict, inputs::Dict)
 
     # Hourly Matching Requirement
     if setup["HourlyMatchingRequirement"] == 1
-        if setup["Benders"] == 1
-            hourly_matching_subperiod!(EP, inputs)
-        else
-            hourly_matching!(EP, inputs)
-        end
+        hourly_matching!(EP, inputs)
```
lbonaldo, line 374: We also need to double check the hourly matching for Benders.

src/model/generate_model.jl
```diff
@@ -327,8 +329,9 @@ function operation_model!(EP::Model, setup::Dict, inputs::Dict)
     end
 
     # Model constraints, variables, expressions related to the co-located VRE-storage resources
+    # (Benders case with VRE_STOR already errored at generate_model entry point)
     if !isempty(inputs["VRE_STOR"])
-        if setup["LDES_Feasible"] == 0 && ((setup["Benders"] == 1 && haskey(inputs, "SubPeriod_Index") && !isempty(inputs["VS_LDS"])) || ((inputs["REP_PERIOD"] > 1) && !isempty(inputs["VS_LDS"])))
+        if (setup["Benders"] == 1 && haskey(inputs, "SubPeriod_Index") && !isempty(inputs["VS_LDS"])) || (inputs["REP_PERIOD"] > 1 && !isempty(inputs["VS_LDS"]))
```
lbonaldo, line 334: Sorry, can you please remind me why we are removing this part here `setup["LDES_Feasible"] == 0`?

lbonaldo (commented):
Hey @dlcole3, again, thank you so much for this additional PR! I added a few comments, most of them are probably related to the merge with the previous PR. Thank you for preparing them one next to the other. Let me know if you have any questions!

example_systems/2_three_zones_w_electrolyzer_and_hourly_matching/Run.jl
dlcole3: Yep, I missed this. There was also a call to "Revise" in the example 12 that I missed.

src/case_runners/case_runner.jl
dlcole3: RunTransportModel is a solution method for DCOPF=1, so this was intended to flag this if the user was trying to set RunTransportModel it but forgot to set DCOPF=1. This warn statement was unclear. I rephrased it to be: 

```julia
@warn "DC_OPF = $(mysetup["DC_OPF"]) and `RunTransportModel` = true. `RunTransportModel`` is a setting for solving with Benders when DCOPF = 1. Did the user mean to set DC_OPF to 1?"
```

src/case_runners/case_runner.jl
dlcole3: Yep, I missed this on the rebase. I deleted these lines.

src/configure_solver/configure_benders.jl
dlcole3: Yep. This also got missed in my rebase. Thanks for catching

src/model/resources/vre_stor/vre_stor.jl
dlcole3: No, I think this was a rebase issue. This line should be removed. There is a guard on VS_ASYM, but then the constraint is on VS_SYM_AC, and is a duplicate of an above call.

src/model/resources/vre_stor/vre_stor.jl
dlcole3: This was an old docstring. I went back to the Benders doc string.

src/model/generate_model.jl
dlcole3: Yep, this was also a rebase issue I think. I have corrected so it calls them properly.

src/model/generate_model.jl
dlcole3: No, another rebase issue. These should be removed because we call it in the operational model, not in generate_model.

src/model/generate_model.jl
dlcole3: I don't think they should be removed. I think this was another rebasing miss on my part. It is corrected now.

src/model/core/transmission/investment_transmission.jl
dlcole3: I think this should be in the capacity_decision.jl file because these need to be added to the operational period. However, in the capacity_decisions, I think these should not be binary (and the integer vCAP and vRETCAP should not be integer). I have deleted the integer and binary setting in the subproblems since they only need to be integer/binary in the master.

src/model/core/discharge/investment_discharge.jl
dlcole3: Yes, and I believe it should be so that the master problem solution is fixed in the subproblem and is treated properly in the subproblem (e.g., in the master, if it is an integer decision due to its capacity_size, then it should be treated similarly in the subproblem).

docs/src/User_Guide/benders_decomposition.md
```diff
@@ -1,11 +1,11 @@
...
+Benders decomposition is accessed by setting `Benders: 1` in the `genx_settings.yml` file. In addition, the user must  planning problem and subproblem solver settings as `[solver_name]_benders_planning_settings.yml` and `[solver_name]__benders_subprob_settings.yml` as well as a settings files for Benders called `benders_settings.yml`. Internally, GenX calls [MacroEnergySolvers.jl](https://github.com/macroenergy/MacroEnergySolvers.jl/tree/main) to run the Benders algorithm. The settings which can be passed to the `benders_settings.yml` file are shown in the following table and are frequently equivalent to the MacroEnergySolvers settings.
```
lbonaldo, line 8:
```suggestion
Benders decomposition is accessed by setting `Benders: 1` in the `genx_settings.yml` file. In addition, the user must  planning problem and subproblem solver settings as `[solver_name]_benders_planning_settings.yml` and `[solver_name]_benders_subprob_settings.yml` as well as a settings files for Benders called `benders_settings.yml`. Internally, GenX calls [MacroEnergySolvers.jl](https://github.com/macroenergy/MacroEnergySolvers.jl/tree/main) to run the Benders algorithm. The settings which can be passed to the `benders_settings.yml` file are shown in the following table and are frequently equivalent to the MacroEnergySolvers settings. 
```

